PR #1, opened 2019-05-16
Add initial structure for generic chart

kahootali:


stakater-user:

PR #2, opened 2019-05-20
allow multiple configmaps & secrets

kahootali:


stakater-user:

stakater-user:

PR #3, opened 2019-05-28
Fix default service annotations for xposer

hamza3202:


stakater-user:
@hamza3202 Chart: application Dry Run successful

PR #5, opened 2019-08-22
remove image tag and make a single image value

kahootali:


stakater-user:

PR #13, opened 2020-02-17
[add-pod-labels] update manifest

aliartiza75:


stakater-user:
@aliartiza75 Chart: application Dry Run successful

aliartiza75:
we will use podLabels instead of additionalPodLabels.

PR #35, opened 2020-06-11
remove unnecessary labels and other doc fix

usamaahmadkhan:
Don't need unnecessary labels. use already defined label `app`


stakater-user:
@usamaahmadkhan Chart: application Dry Run successful

stakater-user:
@usamaahmadkhan Chart: application Dry Run successful

PR #43, opened 2020-09-07
Add volume mounts persistence and fix labels

kahootali:
This PR
- removes default values for rolling update as cant set strategy to recreate due to it
- changes pod label as application label is part of chart now
- adds values for mount pvc and path in case persistence is true


stakater-user:

rasheedamir:
@kahootali update readme as well

stakater-user:

stakater-user:

ahmedwaleedmalik (approved):
LGTM

PR #77, opened 2021-03-26
Allow to append namesuffix to envFrom for configmap and secret

faizanahmad055:


stakater-user:
@faizanahmad055 Validation Successful

usamaahmadkhan:
Plz Add the suffix option to configmap.yaml and secret.yaml aswell. So that if suffix is enabled, the configmap/secret created also have that suffix.

faizanahmad055:
> Plz Add the suffix option to configmap.yaml and secret.yaml aswell. So that if suffix is enabled, the configmap/secret created also have that suffix.

Secret and Configmap already have namesuffix. Is this what you are asking about? [Secret](https://github.com/stakater-charts/application/blob/master/application/templates/secret.yaml#L20) and [values](https://github.com/stakater-charts/application/blob/master/application/values.yaml#L391).

usamaahmadkhan (approved):

PR #94, opened 2021-09-02
fix actions

Nebojsa92:
Fix `push` and `pull_request` actions. Changes

- `helm install ${CHART_NAME} ${CHART_NAME} `**`-f ${CHART_NAME}/values-test.yaml`**` --dry-run --debug`

Fix ingress template:

-  There is no `serviceName` and `servicePort`, but `service.name` and `service.port.name/number`
- Annotations in `values-test.yaml` should have values as strings, even for booleans, eg. `true` should be `"true"`


stakater-user:
@Nebojsa92 Validation Successful

hanzala1234 (approved):

PR #105, opened 2021-09-20
Fix git commit path in github actions

hanzala1234:


stakater-user:

PR #106, opened 2021-09-20
FIx git commit steps

hanzala1234:


stakater-user:

PR #108, opened 2021-09-21
squash noisy commits

Nebojsa92:


stakater-user:
@Nebojsa92 Yikes! You better fix it before anyone else finds out! [Build](https://github.com/stakater-charts/application/commit/85b6929a42e6abbacdaa62c556a25b29265030f0/checks) has Failed!

PR #118, opened 2021-12-10
Delete clusterrole and clusterrolebinding

rasheedamir:
doesn't make sense to create cr and crb with application chart


stakater-user:
@rasheedamir Validation Successful

hanzala1234 (approved):

PR #123, opened 2022-01-10
Add mutual tls for Java based applications

faizanahmad055:

### DIFF
--- a/application/templates/certificate.yaml
+++ b/application/templates/certificate.yaml
@@ -1,5 +1,5 @@
-{{- if .Values.certificate }}
-{{- if .Values.certificate.enabled }}
+{{- if or (.Values.certificate) (.Values.mutualTls.enabled) }}
+{{- if or (.Values.certificate.enabled) (.Values.mutualTls.enabled) }}
 apiVersion: cert-manager.io/v1alpha3
 kind: Certificate
 metadata:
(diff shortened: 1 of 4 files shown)


stakater-user:
@faizanahmad055 Yikes! You better fix it before anyone else finds out! [Build](https://github.com/stakater-charts/application/commit/8bcc3cbcb70b4e2a99b721bf02b95e0457f252c6/checks) has Failed!

stakater-user:
@faizanahmad055 Yikes! You better fix it before anyone else finds out! [Build](https://github.com/stakater-charts/application/commit/be4fd3fee26bd3e60f11918527b16e9f568538da/checks) has Failed!

application/templates/service.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.service }}
+{{- if or (.Values.service) (.Values.mutualTls.enabled) }}
```
hanzala1234, line 1: why is that so?

application/templates/certificate.yaml
```diff
@@ -18,12 +18,20 @@ metadata:
   name: {{ template "application.name" . }}-certificate
   namespace: {{ template "application.namespace" . }}
 spec:
+{{- if .Values.certificate.secretName }}
   secretName: {{ .Values.certificate.secretName }}
+{{- else }}
+  secretName: {{ template "application.name" . }}-client-tls
```
hanzala1234, line 24: also I would suggest to take these in helpers.tpl. we just specify "certificate.secretName" here and conditions come from helpers.tpl. will be helpful to keep templates clean

application/templates/service.yaml
hanzala1234: we should not enable it when TLS is enabled. its like a requirement that in order to enable mTLS, service has to be enabled but this logic does not sound logical here

rasheedamir:
closing as we are going to move this to umbrella chart and not in the leader chart

PR #125, opened 2022-01-18
Add external secrets in application chart

bilalbokharee:


stakater-user:

rasheedamir:
@bilalbokharee data values-test.yaml so, pipeline can validate creation of this resource

application/values.yaml
```diff
@@ -634,3 +634,44 @@ prometheusRule:
     #       for: 1m
     #       labels:
     #         severity: critical  
+##########################################################
+# External Secrets
+##########################################################
+externalSecret:
```
rasheedamir, line 640: so, we are taking input a list or a map? can a value in one environment be overridden by a value in another environment?

stakater-user:
@bilalbokharee Yikes! You better fix it before anyone else finds out! [Build](https://github.com/stakater-charts/application/commit/02fa2349e063a724387787ece9a6bc617a65d69f/checks) has Failed!

stakater-user:
@bilalbokharee Yikes! You better fix it before anyone else finds out! [Build](https://github.com/stakater-charts/application/commit/3d53dc18ef083ad9194663dc6ae63358f705465d/checks) has Failed!

stakater-user:

application/values.yaml
hanzala1234: we are taking as map now. it can be overridden

hanzala1234 (approved):

PR #139, opened 2022-03-04
truncate appVersion to 63 characters

bilalbokharee:


stakater-user:
@bilalbokharee Yikes! You better fix it before anyone else finds out! [Build](https://github.com/stakater-charts/application/commit/c4ba62296c7be493a37a8e9d8562fa7dceade05d/checks) has Failed!

application/templates/_helpers.tpl
```diff
@@ -12,7 +12,7 @@ app: {{ template "application.name" . }}
 
 {{- define "application.labels.stakater" -}}
 {{ template "application.labels.selector" . }}
-appVersion: "{{ .Values.deployment.image.tag }}"
+appVersion: "{{ .Values.deployment.image.tag | trunc 63 | trimSuffix "-" -}}"
```
rasheedamir, line 15: why truncate at 63 chars?

stakater-user:

application/templates/ingress.yaml
```diff
@@ -23,6 +23,7 @@ spec:
       http:
         paths:
           - path: /
+            pathType: Prefix
```
rasheedamir, line 26: @bilalbokharee why Prefix?

hanzala1234, line 26: it can't be hardcoded. take it from values.yaml

application/templates/_helpers.tpl
bilalbokharee: because some Kubernetes name fields are limited to this.

hanzala1234 (approved):

PR #169, opened 2022-05-06
Add support for Deployment startup probe

softstuff:
This probe is good for slow starting applications instead of relying om initialDelaySeconds that will make the pod start slower then necessar.

Docs: 
StartupProbe indicates that the Pod has successfully initialized. If specified, no other probes are executed until this completes successfully. If this probe fails, the Pod will be restarted, just as if the livenessProbe failed. This can be used to provide different probe parameters at the beginning of a Pod's lifecycle, when it might take a long time to load data or warm a cache, than during steady-state operation. This cannot be updated. More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,15 +103,15 @@ To uninstall the chart:
 ##### Startup Probe
 StartupProbe indicates that the Pod has successfully initialized. If specified, no other probes are executed until this completes successfully.
 
-| Name                     | Description                                                                                  | Value           |
-| ------------------------ | -------------------------------------------------------------------------------------------- | --------------- |
-| deployment.startupProbe.enabled | Enabled startup probe                                                                  | true       |
-| deployment.startupProbe.failureThreshold | When a probe fails, Kubernetes will try failureThreshold times before giving up.                                                                  | 3      |
-| deployment.startupProbe.periodSeconds | Perform probe  everytime after specified periodSeconds                                                                  | 10       |
-| deployment.startupProbe.successThreshold | Minimum consecutive successes for the probe to be considered successful after having failed.                                                                  | 1       |
-| deployment.startupProbe.timeoutSeconds | Number of seconds after which the probe times out.                                                                  | 1       |
-| deployment.startupProbe.httpGet | Describes an action based on HTTP Get requests                                                                  |   path: '/path' port: 8080     |
-| deployment.startupProbe.exec | Kubelet executes the specified command to perform the probe                                                                  |   {}   |
+| Name                     | Description                                                                                 | Value                  |
+| ------------------------ |---------------------------------------------------------------------------------------------|------------------------|
+| deployment.startupProbe.enabled | Enabled startup probe                                                                       | false                  |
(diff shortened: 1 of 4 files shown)


stakater-user:
@softstuff Yikes! You better fix it before anyone else finds out! [Build](https://github.com/stakater-charts/application/commit/044e8162fedf2e632e3e9e221d611f75c31d4ed0/checks) has Failed!

stakater-user:
@softstuff Validation Successful

README.md
```diff
@@ -100,7 +100,22 @@ To uninstall the chart:
 
 #### Deployment Probes Paramaters
 
+##### Startup Probe
+StartupProbe indicates that the Pod has successfully initialized. If specified, no other probes are executed until this completes successfully.
+
+| Name                     | Description                                                                                  | Value           |
+| ------------------------ | -------------------------------------------------------------------------------------------- | --------------- |
+| deployment.startupProbe.enabled | Enabled startup probe                                                                  | true       |
```
hanzala1234, line 108: I think be default it should be false. In most of the applications, startupProbes wont be used as liveness probes. it good to have but should not be enforced by default. user will have to provide extra parameters for its configuration

README.md
```diff
@@ -100,7 +100,22 @@ To uninstall the chart:
...
+| deployment.startupProbe.failureThreshold | When a probe fails, Kubernetes will try failureThreshold times before giving up.                                                                  | 3      |
```
hanzala1234, line 109: this value should be bit higher. thats where this probe delivers value that  it does not fail immediately during startup. I would say 10

application/values-test.yaml
```diff
@@ -135,7 +135,16 @@ deployment:
     tag: v1.0.0
     pullPolicy: IfNotPresent
 
-  # Readiness and Liveness probes
+  # Startup, Readiness and Liveness probes
+  startupProbe:
+    enabled: false
```
hanzala1234, line 140: should be true here. as these values are for validation in pipeline

README.md
hanzala1234: seems like its false in values.yaml, just in README its wrong

hanzala1234:
@softstuff  Thanks for PR, I have added some comments.

README.md
softstuff: thanks, just updated the README to reflect the default values. enable is false, successThreshold and timeoutSeconds is not set use k8s default.

README.md
softstuff: I had it bumped 30 (5 min) in the readme as it was that in the values file. Is 30 ok or do you like me to set it to 10 (100 sek)?

application/values-test.yaml
softstuff: fixed

README.md
hanzala1234: 30 should be fine

hanzala1234 (approved):

PR #170, opened 2022-05-09
Remove application.labels.chart from labels

mustafaStakater:
Remove application.labels.chart from labels


stakater-user:

PR #171, opened 2022-05-10
Remove label from application.labels.chart

mustafaStakater:
Remove label from application.labels.chart


stakater-user:

PR #178, opened 2022-05-14
[add-type-in-service]

aliartiza75:
add type in the service template

### DIFF
--- a/application/templates/service.yaml
+++ b/application/templates/service.yaml
@@ -16,6 +16,9 @@ metadata:
   name: {{ template "application.name" . }}
   namespace: {{ template "application.namespace" $ }}
 spec:
+{{- if .Values.service.type }}  
+  type: "{{ .Values.service.type }}"
+{{- end }}
   selector:
 {{ include "application.labels.selector" . | indent 4 }}
 {{- if .Values.deployment.podLabels }}
--- a/README.md
+++ b/README.md
@@ -581,7 +581,6 @@ All notable changes to this project will be documented here
 ### v1.1.14
 - Feature: replica field is made optional
 
-
 ### v1.1.13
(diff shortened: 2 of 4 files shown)


stakater-user:
@aliartiza75 Validation Successful

application/templates/service.yaml
```diff
@@ -16,6 +16,9 @@ metadata:
   name: {{ template "application.name" . }}
   namespace: {{ template "application.namespace" $ }}
 spec:
+{{- if .Values.service.type }}
```
hanzala1234, line 19: I think this condition is not required. As we are providing default value in pipeline. if we dont provide it will still be ClusterIP. 
in short, I dont think anyone will explicitly remove this field from default value

rasheedamir: @aliartiza75 ^^

aliartiza75: @hanzala1234 I agree but it would be good just to have the ability to configure it.

stakater-user:
@aliartiza75 Yikes! You better fix it before anyone else finds out! [Build](https://github.com/stakater-charts/application/commit/90397aa96f0d519956b9b979f1a7bd681574d5ca/checks) has Failed!

rasheedamir:
@aliartiza75 can you plz fix the build?

stakater-user:
@hanzala1234 Yikes! You better fix it before anyone else finds out! [Build](https://github.com/stakater-charts/application/commit/90397aa96f0d519956b9b979f1a7bd681574d5ca/checks) has Failed!

aliartiza75:
@rasheedamir i think the issue is not with the manifest but the pipeline itself.

stakater-user:
@mustafaStakater Yikes! You better fix it before anyone else finds out! [Build](https://github.com/stakater-charts/application/commit/48a28e36f131b83cf6ddcc19f62ce724db936984/checks) has Failed!

stakater-user:

hanzala1234 (approved):

stakater-user:

PR #179, opened 2022-05-17
Change gh-action-push to repush in case of failure

mustafaStakater:
Change gh-action-push to repush in case of failure, it repull from remote and pushes again.

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -91,7 +91,7 @@ jobs:
         git add application/Chart.yaml
         git commit -m "[skip-ci] Update artifacts" 
 
-    # Push Chart.yml with Updated Version
+    # Push Chart.yaml with Updated Version
     - name: Push changes
       shell: bash {0}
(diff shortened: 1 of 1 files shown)


stakater-user:

.github/workflows/push.yaml
```diff
@@ -87,11 +91,40 @@ jobs:
...
 
+    # Push Chart.yml with Updated Version
     - name: Push changes
-      uses: ad-m/github-push-action@master
-      with:
-        github_token: ${{ secrets.STAKATER_GITHUB_TOKEN }}
-
+      shell: bash {0}
+      run: |
+        git config --local user.email "stakater@gmail.com"
```
hanzala1234, line 98: This should come from secret or variable. is it hard coded in other place?

mustafaStakater: yes its hard coded in application chart,

hanzala1234 (approved):

PR #196, opened 2022-10-19
Update ingress to allow multiple path

tianomagdaong:
Modified ingress.yaml to allow multiple paths and remove hardcoded paths.


stakater-user:
@tianomagdaong Yikes! You better fix it before anyone else finds out! [Build](https://github.com/stakater-charts/application/commit/9bc08f4f78d93cfd6b791267fc0e0efbc463f3d9/checks) has Failed!

PR #203, opened 2022-11-23
Enable default service account

faizanahmad055:

### DIFF
--- a/application/values.yaml
+++ b/application/values.yaml
@@ -95,7 +95,7 @@ deployment:
     #   timeFormat: world
 
   # Replicas to be created
-  replicas: 
+  replicas: 1
 
   # Secrets used to pull image
   imagePullSecrets: ""
@@ -182,7 +182,7 @@ deployment:
     exec: {}
 
   readinessProbe:
-    enabled: true
+    enabled: false
     failureThreshold: 3
     periodSeconds: 10
     successThreshold: 1
@@ -194,7 +194,7 @@ deployment:
     exec: {}
 
   livenessProbe:
-    enabled: true
+    enabled: false
     failureThreshold: 3
     periodSeconds: 10
     successThreshold: 1
@@ -429,8 +429,8 @@ rbac:
 
   # Service Account to use by pods
   serviceAccount:
-    enabled: false
-    name: ""
+    enabled: true
+    name: "application"
 
     # Additional Labels on service account
     additionalLabels:


stakater-user:
@faizanahmad055 Yikes! You better fix it before anyone else finds out! [Build](https://github.com/stakater/application/commit/1b2d2afa08d59c0dc587905bfb1606da301ef092/checks) has Failed!

TehreemNisa (approved):

stakater-user:
@faizanahmad055 Validation Successful

application/values.yaml
```diff
@@ -95,7 +95,7 @@ deployment:
     #   timeFormat: world
 
   # Replicas to be created
-  replicas: 
+  replicas: 1
```
callum-stakater, line 98: This doesnt help and will causes issues when used in conjunction with HorizontalPodAutoscaling 

For example ArgoCD will sync 1 replica, HPA will try scale up and they will flip flop for ever

callum-stakater (changes requested):
see comments

application/values.yaml
rasheedamir: we should add this in the values file; I mean this reasoning

application/values.yaml
```diff
@@ -429,8 +429,8 @@ rbac:
 
   # Service Account to use by pods
   serviceAccount:
-    enabled: false
-    name: ""
+    enabled: true
+    name: "application"
```
rasheedamir, line 433: @faizanahmad055 hardcoding to **application**? shouldn't it be a template? what if we deploy five applications in same namespace then they will use the same service account

rasheedamir: https://github.com/stakater/application/blob/master/application/templates/serviceaccount.yaml#L24

rasheedamir: in this case we should just remove name from the values.yaml and then it will automatically get the templatized name

application/values.yaml
```diff
@@ -182,7 +182,7 @@ deployment:
     exec: {}
 
   readinessProbe:
-    enabled: true
+    enabled: false
```
callum-stakater, line 185: These should be carefully considered based on the purpose of this chart which is to enforce best practices not as a way to have a 3 line deployment

Best practice is that all deployments should have probes and the current config forces a developer to define their probes or explicitly disable them rather than defaulting to no probes

application/values.yaml
```diff
@@ -429,8 +429,8 @@ rbac:
 
   # Service Account to use by pods
   serviceAccount:
-    enabled: false
-    name: ""
+    enabled: true
```
callum-stakater, line 432: The purpose of having SA enabled by default is again a best practice by default thing, best practice is that every deployment should bring its own SA and  not mount the default SA's

application/values.yaml
callum-stakater: also just to point out a deployment that omits the `replicas:` is still perfectly valid, kubernetes defaults to 1 replica

application/values.yaml
rasheedamir: this should be taken from **helpers.tpl**

mustafaStakater:
We can close this pr

PR #208, opened 2022-12-01
Disable Deployment Probes

mustafaStakater:
Comment Deployment Probes.  liveness probe in default values is overriding liveness probe in values outside because values from default are merged with values from outside causing both tcpSocket and httpGet to be present, chart gives precendence to exec over httpGet over tcpSocket, this is causing the pod to be stuck in a crashback loop because probe will be different for every pod

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 
 All notable changes to this project will be documented here
 
-### v1.3.1
+### v1.3.2
 - Feature: Disable liveness and readiness probes.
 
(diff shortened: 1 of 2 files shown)


stakater-user:
@mustafaStakater Yikes! You better fix it before anyone else finds out! [Build](https://github.com/stakater/application/commit/5f38c7419b9247c755ab7fe6fc20860ee4f56314/checks) has Failed!

AsfaMumtaz (approved):

Issue #195, opened 2022-10-13
Expecting Service Account when disabled.


amanhigh:
Default Application doesn't run until Service Account is created.

```
applicationName: "tiny"
deployment:
  image:
    repository: travix/tinyproxy
    tag: latest
```

The above config kept on failing at replica set level. 

Issues making it difficult to use in start.
- No error when deploying, but pods won't come up.
- Not a single example on Homepage which can be copy pasted and run. Eg. (Sample Nginx)

amanhigh:
To make it work i had to give following config.


`##########################################################
# Name of the application.
##########################################################
applicationName: "tiny"

##########################################################
# Deployment
##########################################################
deployment:
  # Image of the app container
  image:
    repository: travix/tinyproxy
    tag: latest

  replicas: 1

  readinessProbe:
    enabled: false
  livenessProbe:
    enabled: false

# Service Account to use by pods
rbac:
  serviceAccount:
    enabled: true
    name: "tiny"
`

rasheedamir:
good points @amanhigh would you like to contribute a PR? and add minimalistic example on the home page as well?

PR #213, opened 2023-01-16
fix: optional service account

rasheedamir:
don't add service account until user asks for it! it will get default service account
Fixes https://github.com/stakater/application/issues/195

### DIFF
--- a/application/templates/deployment.yaml
+++ b/application/templates/deployment.yaml
@@ -276,10 +276,12 @@ spec:
         {{- end }}
         {{- end }}
       {{- end }}
+      {{- if .Values.rbac }}
+      {{- if .Values.rbac.serviceAccount }}
       {{- if .Values.rbac.serviceAccount.name }}
       serviceAccountName: {{ .Values.rbac.serviceAccount.name }}
-        {{- else }}
-      serviceAccountName: {{ template "application.name" . }}
       {{- end }}
+      {{- end }}
+      {{- end }}      
 {{- end }}   
 {{- end }}
--- a/application/templates/deployment.yaml
+++ b/application/templates/deployment.yaml
@@ -283,9 +283,5 @@ spec:
       serviceAccountName: {{ template "application.name" $ }}
       {{- end }}
       {{- end }}
-      serviceAccountName: {{ .Values.rbac.serviceAccount.name }}
-      {{- end }}
-      {{- end }}
-      {{- end }}      
 {{- end }}   
 {{- end }}
--- a/application/templates/deployment.yaml
+++ b/application/templates/deployment.yaml
@@ -276,9 +276,13 @@ spec:
         {{- end }}
         {{- end }}
       {{- end }}
-      {{- if .Values.rbac }}
-      {{- if .Values.rbac.serviceAccount }}
+      {{- if .Values.rbac.serviceAccount.enabled }}
       {{- if .Values.rbac.serviceAccount.name }}
+      serviceAccountName: {{ .Values.rbac.serviceAccount.name }}
+      {{- else }}
+      serviceAccountName: {{ template "application.name" $ }}
+      {{- end }}
+      {{- end }}
       serviceAccountName: {{ .Values.rbac.serviceAccount.name }}
       {{- end }}
       {{- end }}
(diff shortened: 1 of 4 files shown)


stakater-user:
@rasheedamir Validation Successful

rasheedamir:
@AsfaMumtaz plz update the changelog as well

application/values-test.yaml
```diff
@@ -386,7 +386,7 @@ rbac:
   # Service Account to use by pods
   serviceAccount:
     enabled: true
-    name: ""
+    name: "application"
```
rasheedamir, line 389: this shouldn't be hardcoded instead generated dynamically

application/templates/deployment.yaml
```diff
@@ -276,10 +276,12 @@ spec:
         {{- end }}
         {{- end }}
       {{- end }}
+      {{- if .Values.rbac }}
+      {{- if .Values.rbac.serviceAccount }}
       {{- if .Values.rbac.serviceAccount.name }}
```
d3adb5, line 281: I think we can leverage order of evaluation here to simplify this line. This seems to be working as of Helm v3.11.1:

```suggestion
      {{- if and .Values.rbac .Values.rbac.serviceAccount .Values.rbac.serviceAccount.name }}
```

However I think we should instead just make sure `.Values.rbac.serviceAccount` is at least `{}` in our `values.yaml`, **which we already do!** So I recommend going simply with:

```suggestion
      {{- with .Values.rbac.serviceAccount.name }}
```

And using `.` within. This is because to break the above line, one would have to provide overrides like:

```yaml
rbac:
```

or

```yaml
rbac:
  serviceAccount:
```

At that point, I'd expect the chart user to know what they're doing and understand what to do when template rendering fails.

application/values-test.yaml
d3adb5: These values are used to ensure the templates are rendered properly during a `helm install`, right? What's your idea for generating this value dynamically? :thinking:

d3adb5 (changes requested):

unixfox:
Looking forward for this PR because currently it's not possible to use this chart if you don't want to have any RBAC involved.

application/templates/deployment.yaml
mustafaStakater: We need to define default service account `{{ template "application.name" $ }}` in else block incase `.Values.rbac.serviceAccount.name` isnt defined, using `if` is fine

mustafaStakater: What do you think about going forward as below

```suggestion
      {{- if .Values.rbac.serviceAccount.enabled }}
      {{- if $.Values.rbac.serviceAccount.name }}
      serviceAccountName: {{ $.Values.rbac.serviceAccount.name }}
      {{- else }}
      serviceAccountName: {{ template "application.name" $ }}
      {{- end }}
      {{- end }}
```

aslafy-z (approved):

application/templates/deployment.yaml
```diff
@@ -291,9 +291,11 @@ spec:
...
       {{- end }}
+      {{- if .Values.rbac.serviceAccount.enabled }}
       {{- if .Values.rbac.serviceAccount.name }}
       serviceAccountName: {{ .Values.rbac.serviceAccount.name }}
-        {{- else }}
-      serviceAccountName: {{ template "application.name" . }}
+      {{- else }}
+      serviceAccountName: {{ template "application.name" $ }}
+      {{- end }}
       {{- end }}
```
d3adb5, line 300: I think it's worth adding a chart unit test to check that the `serviceAccountName` field is empty. We have three possible branches here: populated and proxied from values, populated via the fallback (`template "application.name" $`), and empty.

At the very least we should test that it's empty when disabled:

```yaml
- it: yields empty service account name when disabled
  set:
    rbac.serviceAccount.enabled: false
  asserts:
    - isEmpty:
        path: spec.template.spec.serviceAccountName
```

d3adb5 (changes requested):

application/templates/deployment.yaml
aslafy-z: I added tests to cover the three branches. Let me know if it fits you :)

rasheedamir:
@d3adb5 is it good to merge now?

aslafy-z:
@rasheedamir I would like to wait for @d3adb5 approval first, but yes, I think we're good to go.

d3adb5 (approved):
LGTM!

PR #221, opened 2023-02-20
migrate cronjob and hpa to new api versions

pauvos:
* Cronjob v1beta1 => v1
* HorizontalPodAutoscaler v2beta2 => v2

see https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-25

### DIFF
--- a/application/templates/hpa.yaml
+++ b/application/templates/hpa.yaml
@@ -1,5 +1,5 @@
 {{- if and (.Values.autoscaling).enabled .Values.deployment }}
-{{ if $.Capabilities.APIVersions.Has "autoscaling/v2/HorizontalPodAutoscaler" -}}
+{{ if .Capabilities.APIVersions.Has "autoscaling/v2/HorizontalPodAutoscaler" -}}
 apiVersion: autoscaling/v2
 {{- else -}}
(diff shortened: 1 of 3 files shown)


stakater-user:
@pauvos Yikes! You better fix it before anyone else finds out! [Build](https://github.com/stakater/application/commit/fe536f76c4f5dc84c65a16f9756084f18d452989/checks) has Failed!

rasheedamir:
@pauvos can you plz update this?

d3adb5 (approved):

d3adb5:
Looks good to me, but since I'm a new maintainer, I'd rather have another maintainer take a look and let me know if I missed something like a necessary addition to the CHANGELOG or something similar.

application/templates/hpa.yaml
```diff
@@ -1,5 +1,9 @@
 {{- if and (.Values.autoscaling).enabled .Values.deployment }}
+{{ if $.Capabilities.APIVersions.Has "autoscaling/v2/HorizontalPodAutoscaler" -}}
```
mustafaStakater, line 2: Why use $ when youre not within `with` or `range`

mustafaStakater (approved):

Issue #225, opened 2023-03-13
Add .Capabilities.APIVersions.Has on non k8s core resources


rasheedamir:
We should add `.Capabilities.APIVersions.Has` to non k8s core resources

rasheedamir:
will be fixed by https://github.com/stakater/application/pull/226

Issue #224, opened 2023-03-13
Add support for namespaceOverride


rasheedamir:
There should be possibility to override namespace like this

```
{{/*
Allow the release namespace to be overridden
*/}}
{{- define "rabbitmq-operator.namespace" -}}
{{- default .Release.Namespace .Values.namespaceOverride -}}
{{- end -}}
```

rasheedamir:
will be fixed by https://github.com/stakater/application/pull/226

Issue #202, opened 2022-11-23
Review the labels and remove the unncessary ones


rasheedamir:
We should review these labels; as some of them are static and don't make sense

```
  labels:
    app: review
    app.kubernetes.io/instance: gabbar-dev-stakater-nordmart-review
    release: gabbar-dev-stakater-nordmart-review
    provider: stakater
    appVersion: 1.0.43
    team: stakater
    chart: application-1.1.14
    heritage: Helm
    appCategory: backend
    group: com.stakater.platform
```

We can remove these:
- provider: stakater
- team: stakater

And have different names for others

rasheedamir:
will be fixed by https://github.com/stakater/application/pull/226

Issue #201, opened 2022-11-23
Add the common k8s recommended common labels


rasheedamir:
Add these common recommended common labels

```
    app.kubernetes.io/name: mysql
    app.kubernetes.io/instance: mysql-abcxzy
    app.kubernetes.io/version: "5.7.21"
    app.kubernetes.io/component: database
    app.kubernetes.io/part-of: wordpress
    app.kubernetes.io/managed-by: helm
```

https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

rasheedamir:
will be fixed by https://github.com/stakater/application/pull/226

PR #226, opened 2023-03-13
improve enable conditions and use helpers.tpl

rasheedamir:
- improve enable conditions
- user _helpers.tpl properly
- move to common k8s labels and remove other labels

And fixes following:

- [ ] #201 
- [ ] #202 
- [ ] #224 
- [ ] #225

### DIFF
--- a/application/templates/_helpers.tpl
+++ b/application/templates/_helpers.tpl
@@ -8,6 +8,13 @@ Define the name of the chart/application.
 {{- end -}}
 
 {{/*
+Define the name of the chart/application.
+*/}}
+{{- define "application.version" -}}
+{{ regexReplaceAll "[^a-zA-Z0-9_\\.\\-]" .Values.deployment.image.tag "-" | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
 Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 If release name contains chart name it will be used as a full name.
@@ -51,7 +58,7 @@ Common labels
 {{- define "application.labels" -}}
 helm.sh/chart: {{ include "application.chart" . }}
 app.kubernetes.io/name: {{ include "application.name" . }}
-app.kubernetes.io/version: {{- .Values.deployment.image.tag | trunc 63 | trimSuffix "-" -}}
+app.kubernetes.io/version: {{ include "application.version" . }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 app.kubernetes.io/component: {{ include "application.name" . }}
 app.kubernetes.io/part-of: {{ include "application.name" . }}
--- a/application/templates/_helpers.tpl
+++ b/application/templates/_helpers.tpl
@@ -15,24 +15,6 @@ Define the name of the chart/application.
 {{- end -}}
 
 {{/*
-Create a default fully qualified app name.
-We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
-If release name contains chart name it will be used as a full name.
-*/}}
-{{- define "application.fullname" -}}
-{{- if .Values.fullnameOverride }}
-{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
-{{- else }}
-{{- $name := default .Chart.Name .Values.nameOverride }}
-{{- if contains $name .Release.Name }}
-{{- .Release.Name | trunc 63 | trimSuffix "-" }}
-{{- else }}
-{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
-{{- end }}
-{{- end }}
-{{- end }}
-
-{{/*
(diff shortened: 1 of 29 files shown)


application/templates/_helpers.tpl
```diff
@@ -36,13 +45,35 @@ Usage:
...
+{{- define "application.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Common labels
+*/}}
+{{- define "application.labels" -}}
+helm.sh/chart: {{ include "application.chart" . }}
+app.kubernetes.io/name: {{ include "application.name" . }}
```
mustafaStakater, line 60: this might conflict with argocd, since argocd adds the same label with value corrresponding to app name 
`app.kubernetes.io/name: argocd-application-name`

rasheedamir: it's a good point but the question if one isn't using argocd then what will happen? can you please read and check if this will become an issue with argocd?

application/templates/_helpers.tpl
```diff
@@ -1,28 +1,37 @@
...
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "application.fullname" -}}
+{{- if .Values.fullnameOverride }}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- $name := default .Chart.Name .Values.nameOverride }}
```
mustafaStakater, line 26: why are we defining it, if its not being used by any template

application/templates/_helpers.tpl
mustafaStakater: ok i can take a look into this

application/templates/alertmanagerconfig.yaml
```diff
@@ -1,13 +1,11 @@
...
 apiVersion: monitoring.coreos.com/v1alpha1
 kind: AlertmanagerConfig
 metadata:
   name: {{ template "application.name" . }}
-  namespace: {{ template "application.namespace" . }}
+  namespace: {{ include "application.namespace" . }}
   labels:
-{{ include "application.labels.stakater" . | indent 4 }}
-{{ include "application.labels.chart" . | indent 4 }}
+{{ include "application.labels" . | indent 4 }}
```
mustafaStakater, line 8: we can do `{{- include "application.labels" . | nindent 4 }}`  and indent this line

rasheedamir: we should indent in all files; @mustafaStakater can you do that?

application/templates/pvc.yaml
```diff
@@ -1,24 +1,22 @@
-{{- if .Values.persistence }}
-{{- if and .Values.persistence.enabled (not .Values.persistence.existingClaim) }}
+{{- if and (.Values.persistence).enabled (not .Values.persistence.existingClaim) }}
 kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
+{{- if .Values.persistence.name }}
+  name: "{{ .Values.persistence.name }}"
+{{- else }}
+  name: {{ template "application.name" . }}-data
```
mustafaStakater, line 8: we need to update claim name in deployment.yaml https://github.com/stakater/application/blob/5bc7a996b42206e58c6cd248f916282fa79fa609/application/templates/deployment.yaml#L276

rasheedamir: great catch @mustafaStakater plz fix it

application/templates/alertmanagerconfig.yaml
mustafaStakater: fixed

application/templates/pvc.yaml
mustafaStakater: fixed

application/templates/_helpers.tpl
rasheedamir: @mustafaStakater plz remove if not required

mustafaStakater: removed

application/templates/_helpers.tpl
```diff
@@ -39,10 +39,8 @@ Common labels
 */}}
 {{- define "application.labels" -}}
 helm.sh/chart: {{ include "application.chart" . }}
-app.kubernetes.io/name: {{ include "application.name" . }}
 app.kubernetes.io/version: {{ include "application.version" . }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
-app.kubernetes.io/component: {{ include "application.name" . }}
```
rasheedamir: @mustafaStakater if someone isn't using argocd then what will happen?

mustafaStakater: it shouldnt be a problem in that case, i left part-of label for users to identify what application is this a part of

mustafaStakater (approved):

PR #234, opened 2023-03-24
make deployment.image.tag and job.image.tag optional

slauger:
Hi,

i really like the idea of creating a generic chart which can be used for many use cases, but i have a small issue with the Chart.

Currently you use the the following code block for the image definition:

```
        {{- if .Values.deployment.image.digest }}
        image: "{{ .Values.deployment.image.repository }}@{{ .Values.deployment.image.digest }}"
        {{- else if .Values.deployment.image.tag }}
        image: "{{ .Values.deployment.image.repository }}:{{ .Values.deployment.image.tag }}"
        {{- end }}
```

This means that currently `deployment.image.tag` is a required param. The current default value is `v1.0.0`. This is a bit odd, because a image definition is also valid w/o a tag. For example, the following definitions are also valid and not all of them are currently supported by the chart:

```
quay.io/xyz/foo:v3.0.5
quay.io/xyz/baz:latest
quay.io/xyz/bar@sha256:686d8c9dfa6f3ccfc8230bc3178d23f84eeaf7e457f36f271ab1acc53015037c
quay.io/xyz/foo:latest@sha256:686d8c9dfa6f3ccfc8230bc3178d23f84eeaf7e457f36f271ab1acc53015037c
```

Also it's very common that you receive the image as complete "all-in-one" string from a CI/CD pipeline. Extracting just the digest or tag and hand it over to helm could be very frustrating and complex sometimes. So it would be great if you just set the image with the tag and digest as a single parameter. 

Attached you will find a pull request which makes `deployment.image.tag` (and `job.image.tag`) optional and set the default value to `null`.

Let me know what you think about this change.

### DIFF
--- a/application/templates/_helpers.tpl
+++ b/application/templates/_helpers.tpl
@@ -39,7 +39,9 @@ Common labels
 */}}
 {{- define "application.labels" -}}
 helm.sh/chart: {{ include "application.chart" . }}
+{{- if .Values.deployment.image.tag }}
 app.kubernetes.io/version: {{ include "application.version" . | quote }}
+{{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 app.kubernetes.io/part-of: {{ include "application.name" . }}
 {{- end }}
--- a/application/templates/_helpers.tpl
+++ b/application/templates/_helpers.tpl
@@ -11,7 +11,8 @@ Define the name of the chart/application.
 Define the name of the chart/application.
 */}}
 {{- define "application.version" -}}
-{{ regexReplaceAll "[^a-zA-Z0-9_\\.\\-]" .Values.deployment.image.tag "-" | trunc 63 | trimSuffix "-" -}}
+  {{- $version := default "" .Values.deployment.image.tag -}}
+  {{- regexReplaceAll "[^a-zA-Z0-9_\\.\\-]" $version "-" | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
 {{/*
@@ -39,8 +40,8 @@ Common labels
 */}}
 {{- define "application.labels" -}}
 helm.sh/chart: {{ include "application.chart" . }}
-{{- if .Values.deployment.image.tag }}
-app.kubernetes.io/version: {{ include "application.version" . | quote }}
+{{- with include "application.version" . }}
+app.kubernetes.io/version: {{ quote . }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 app.kubernetes.io/part-of: {{ include "application.name" . }}
--- a/application/templates/cronjob.yaml
+++ b/application/templates/cronjob.yaml
@@ -43,17 +43,12 @@ spec:
           {{- end }}
           containers:
           - name: {{ $name }}
-            {{- if $job.image.tag }}
-            {{- if $job.image.digest }}
-            image: "{{ $job.image.repository }}:{{ $job.image.tag }}@{{ $job.image.digest }}"
-            {{- else }}
-            image: "{{ $job.image.repository }}:{{ $job.image.tag }}"
-            {{- end }}
-            {{- else if $job.image.digest }}
-            image: "{{ $job.image.repository }}@{{ $job.image.digest }}"
-            {{- else }}
-            image: "{{ $job.image.repository }}"
-            {{- end }}
+
+            {{- $image := required (print "Undefined image repo for container '" $name "'") $job.image.repository }}
+            {{- with $job.image.tag    }} {{- $image = print $image ":" . }} {{- end }}
+            {{- with $job.image.digest }} {{- $image = print $image "@" . }} {{- end }}
+            image: {{ $image }}
+
             {{- if $job.image.imagePullPolicy }}
             imagePullPolicy: {{ $job.image.imagePullPolicy }}
             {{ end }}
--- a/application/templates/cronjob.yaml
+++ b/application/templates/cronjob.yaml
@@ -43,10 +43,16 @@ spec:
           {{- end }}
           containers:
           - name: {{ $name }}
+            {{- if $job.image.tag }}
             {{- if $job.image.digest }}
-            image: "{{ $job.image.repository }}@{{ $job.image.digest }}"
-            {{- else if $job.image.tag }}
+            image: "{{ $job.image.repository }}:{{ $job.image.tag }}@{{ $job.image.digest }}"
+            {{- else }}
             image: "{{ $job.image.repository }}:{{ $job.image.tag }}"
+            {{- end }}
+            {{- else if $job.image.digest }}
+            image: "{{ $job.image.repository }}@{{ $job.image.digest }}"
+            {{- else }}
+            image: "{{ $job.image.repository }}"
             {{- end }}
             {{- if $job.image.imagePullPolicy }}
             imagePullPolicy: {{ $job.image.imagePullPolicy }}
--- a/application/templates/deployment.yaml
+++ b/application/templates/deployment.yaml
@@ -130,17 +130,12 @@ spec:
           name: proxy-tls 
         {{- end }}
       - name: {{ template "application.name" . }}
-        {{- if .Values.deployment.image.tag }}
-        {{- if .Values.deployment.image.digest }}
-        image: "{{ .Values.deployment.image.repository }}:{{ .Values.deployment.image.tag }}@{{ .Values.deployment.image.digest }}"
-        {{- else }}
-        image: "{{ .Values.deployment.image.repository }}:{{ .Values.deployment.image.tag }}"
-        {{- end }}
-        {{- else if .Values.deployment.image.digest }}
-        image: "{{ .Values.deployment.image.repository }}@{{ .Values.deployment.image.digest }}"
-        {{- else }}
-        image: "{{ .Values.deployment.image.repository }}"
-        {{- end }}
+
+        {{- $image := required "Undefined image for application container" .Values.deployment.image.repository }}
+        {{- with .Values.deployment.image.tag    }} {{- $image = print $image ":" . }} {{- end }}
+        {{- with .Values.deployment.image.digest }} {{- $image = print $image "@" . }} {{- end }}
+        image: {{ $image }}
+
         imagePullPolicy: {{ .Values.deployment.image.pullPolicy }}
         {{- if .Values.deployment.command }}
         command: {{- include "application.tplvalues.render" (dict "value" .Values.deployment.command "context" $) | nindent 12 }}
--- a/application/templates/deployment.yaml
+++ b/application/templates/deployment.yaml
@@ -130,10 +130,16 @@ spec:
           name: proxy-tls 
         {{- end }}
       - name: {{ template "application.name" . }}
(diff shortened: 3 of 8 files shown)


rasheedamir:
@slauger please update both readme and changelog! thanks for contribution

slauger:
@rasheedamir changelog and readme is updated

application/values.yaml
```diff
@@ -169,7 +169,7 @@ deployment:
   # Image of the app container
   image: 
     repository: repository/image-name
-    tag: v1.0.0
+    tag: null
```
d3adb5, line 172: I personally think an empty string is a nicer default value here, as it is already used elsewhere, is still falsy, and conveys that the expected type is string rather than 'null', which avoids confusion. Moreover, `.Values.deployment.image.tag` could still be passed to template functions expecting strings, like `regexReplaceAll`.

application/templates/cronjob.yaml
```diff
@@ -39,11 +39,17 @@ spec:
...
-            {{- else if $job.image.tag }}
+            image: "{{ $job.image.repository }}:{{ $job.image.tag }}@{{ $job.image.digest }}"
+            {{- else }}
             image: "{{ $job.image.repository }}:{{ $job.image.tag }}"
             {{- end }}
+            {{- else if $job.image.digest }}
+            image: "{{ $job.image.repository }}@{{ $job.image.digest }}"
+            {{- else }}
+            image: "{{ $job.image.repository }}"
+            {{- end }}
```
d3adb5, line 52: To avoid repetition, here's a suggestion:

```gotpl
{{- $image := $job.image.repository }}
{{- with $job.image.tag }}{{ $image = print $image ":" . }}{{ end }}
{{- with $job.image.digest }}{{ $image = print $image "@" . }}{{ end }}
image: {{ $image }}
```

I do believe, however, that if an image repository isn't specified, we should have a templating error, for containers can't be without an image:

```gotpl
{{- $image := $job.image.repository }}
{{- if not $image }}{{ print "Undefined image at CronJob definition at: " $name | fail }}{{ end }}
{{- with $job.image.tag }}{{ $image = print $image ":" . }}{{ end }}
{{- with $job.image.digest }}{{ $image = print $image "@" . }}{{ end }}
image: {{ $image }}
```

If we go this route, there should be a unit test for it!

application/templates/deployment.yaml
```diff
@@ -126,11 +126,17 @@ spec:
...
-        {{- else if .Values.deployment.image.tag }}
+        image: "{{ .Values.deployment.image.repository }}:{{ .Values.deployment.image.tag }}@{{ .Values.deployment.image.digest }}"
+        {{- else }}
         image: "{{ .Values.deployment.image.repository }}:{{ .Values.deployment.image.tag }}"
         {{- end }}
+        {{- else if .Values.deployment.image.digest }}
+        image: "{{ .Values.deployment.image.repository }}@{{ .Values.deployment.image.digest }}"
+        {{- else }}
+        image: "{{ .Values.deployment.image.repository }}"
+        {{- end }}
```
d3adb5, line 139: See other comment about lowering repetition on the CronJob template file.

application/templates/_helpers.tpl
```diff
@@ -39,7 +39,9 @@ Common labels
 */}}
 {{- define "application.labels" -}}
 helm.sh/chart: {{ include "application.chart" . }}
+{{- if .Values.deployment.image.tag }}
 app.kubernetes.io/version: {{ include "application.version" . }}
+{{- end }}
```
d3adb5, line 44: With an empty string as the default value for `.Values.deployment.image.tag`, we could discard this `if` statement, though we would end up with an empty value for a version label. How about leaving out this conditional and instead editing the `application.version` template so that it:

- Tries to use the tag as it currently does, but failing that, it
- Tries to use the digest with a little bit of RegEx substitution magic, but if not provided either, it
- Simply says "latest", as is default behavior for container images?

How does that sound?

application/templates/deployment.yaml
```diff
@@ -291,4 +297,4 @@ spec:
         {{- else }}
       serviceAccountName: {{ template "application.name" . }}
       {{- end }}
-{{- end }}
\ No newline at end of file
+{{- end }}
```
d3adb5: Whitespace changes mess with blame information. Harmless as it is, I'd suggest removing this change any commits that'll be merged.

d3adb5 (dismissed):
Probably completely out of line since I'm not a repository maintainer, but I thought I'd leave a review anyway. Let me know your thoughts, everyone.

application/templates/cronjob.yaml
slauger: It would be possible to remove the default for `image.repository` via

`{{ required "A valid value for .Values.repository.image is required!" .Values.image.repository }}`

but since the default was already set to `repository/image-name` i did not change this behaviour.

application/values.yaml
slauger: Also finds for me. I personally prefer to use null as an undefined value than an empty string.

application/templates/_helpers.tpl
d3adb5: > The application version should always be a semantic version.

I'm not sure if I agree with this one. Helm does require chart versions to be semantic, but the versioning scheme for an application is an incredibly arbitrary choice, usually left to the core maintainers or company behind the project. Hence Helm will let `appVersion` be just about anything in `Chart.yaml` — in fact `.Chart.AppVersion` is often used as the default value when an image tag isn't provided, but this is a generic chart and such a value can't be set during installation.

So that said, whatever logic is present in `application.version`, now or in the future, this could be changed to a simple:

```suggestion
{{- with include "application.version" .  }}
app.kubernetes.io/version: {{ quote . }}
{{- end }}
```

What do you think?

application/templates/cronjob.yaml
d3adb5: Looks like generic default values like those could be refactored out to allow for these safeguards to be put in place. That's probably outside the scope of this pull request, however.

application/templates/_helpers.tpl
slauger: Okay, i see. But what if there is no tag defined? Maybe the tag is a good default for the label, but we could allow overriding the label - maybe with another Value, e.g. `.Values.applicationVersion`? 
Also there could be situations where the version of the application stays the same (e.g. the wordpress version) but the underlying base container receives an update (e.g. packages updates of the alpine/ubi container).

rasheedamir:
@slauger @d3adb5 @aslafy-z I have added all three of you as maintainers of this repo 👍 lets manage it together! We can discuss stuff on slack

slauger:
So how shall we continue here? 

btw why do you maintain the `CHANGELOG.md` manually? You could avoid those merge conflicts easily. The most common tool for releases [semantic-release](https://github.com/semantic-release/github) could help maybe.

slauger:
@rasheedamir ah i see. On which slack workspace can i reach you? I'm already on the Kubernetes slack workspace for example.

application/templates/_helpers.tpl
d3adb5: An `appVersion` value sounds good to me. In the case of no such value and no tag being defined, however, we can just skip the label altogether as it makes no sense to set it, and hence the suggested `with` being appropriate.

d3adb5:
@slauger Do you mind if I push any commits to your branch? Also, there are conflicts now that other PRs have been merged. Let me know what you think of the changes I requested.

rasheedamir:
@slauger can we proceed on this issue and get it done? @d3adb5 plz suggest/push your changes so, we can merge this PR

d3adb5:
I rebased @slauger's branch on master, added tests and did some minor refactoring. I'd appreciate a review since the code did change a little. It's ultimately up to you, @rasheedamir whether we go for another review round or merge this as is.

Strange thing I noticed: the logs for the unittest check show 35 tests, but there are in fact 45 in the latest commit here. Wonder what's up with that. Feels like it's not actually running off the latest commit. :(

rasheedamir:
@mustafaStakater can you plz do the final review of this?

application/templates/cronjob.yaml
```diff
@@ -43,11 +43,12 @@ spec:
           {{- end }}
           containers:
           - name: {{ $name }}
-            {{- if $job.image.digest }}
-            image: "{{ $job.image.repository }}@{{ $job.image.digest }}"
-            {{- else if $job.image.tag }}
-            image: "{{ $job.image.repository }}:{{ $job.image.tag }}"
-            {{- end }}
+
+            {{- $image := required (print "Undefined image repo for container '" $name "'") $job.image.repository }}
```
aslafy-z, line 47:
```suggestion
            {{- $image := required (print "Undefined image repo for cronjob container '" $name "'") $job.image.repository }}
```

application/tests/cronjob_test.yaml
```diff
@@ -0,0 +1,79 @@
...
+        enabled: true
+        jobs:
+          example:
+            image:
+              repository: ""
+              tag: doesnt
+              digest: matter
+    asserts:
+      - failedTemplate:
+          errorMessage: "Undefined image repo for container 'example'"
```
aslafy-z, line 34:
```suggestion
          errorMessage: "Undefined image repo for cronjob container 'example'"
```

application/templates/cronjob.yaml
d3adb5: I find this redundant, as Helm will display the template file where the error was thrown already. Here's an example from a test chart, where I can `helm template --debug .` twice, with errors through in different templates, but with the same error message:

```
install.go:194: [debug] Original chart version: ""
install.go:211: [debug] CHART PATH: /tmp/helm-test


Error: execution error at (test-chart/templates/cronjob.yaml:13:16): This is a failure message
helm.go:84: [debug] execution error at (test-chart/templates/cronjob.yaml:13:16): This is a failure message
```

And then in a Deployment template:

```
install.go:194: [debug] Original chart version: ""
install.go:211: [debug] CHART PATH: /tmp/helm-test


Error: execution error at (test-chart/templates/deployment.yaml:17:12): This is a failure message
helm.go:84: [debug] execution error at (test-chart/templates/deployment.yaml:17:12): This is a failure message
```

What do you think?

application/tests/cronjob_test.yaml
d3adb5: See comment on the other open thread.

application/templates/cronjob.yaml
aslafy-z: I was not sure, looks ok to me :)

aslafy-z:
@d3adb5 the use of `pull_request_target` event cause the CI to run on the base branch context. See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

We might want to reference the pull request head ref in the unittest checkout action so latest code is gathered before executing the actual tests.

d3adb5:
Thank you for the clarification, @aslafy-z! I just submitted a fix in #251 that should be merged before any other PRs, as it could otherwise lead to PRs breaking unit tests being merged.

rasheedamir (approved):

rasheedamir:
LGTM

slauger:
Sorry, i was out of office for a few days. Thanks for taking care of this! ❤️

PR #239, opened 2023-03-27
feat: add topologySpreadConstraints

aslafy-z:
Add values for topologySpreadConstraints


aslafy-z:
@mustafaStakater Can you please have a look? 
Also, this changelog updates and way the releases are done is counterintuitive, I'd like to see semantic-release like automatic versioning in such a simple repository.  I will open an issue about that.

application/templates/deployment.yaml
```diff
@@ -92,6 +92,10 @@ spec:
       {{- if .Values.deployment.affinity }}
       affinity:
 {{ toYaml .Values.deployment.affinity | indent 8 -}}
+      {{- end }}
+      {{- if .Values.deployment.topologySpreadConstraints }}
+      topologySpreadConstraints:
+{{ toYaml .Values.deployment.topologySpreadConstraints | indent 8 }}
       {{- end }}
```
d3adb5, line 99: It would be nice to try and match the enclosing indentation. Purely for readability purposes.

```suggestion
      {{- with .Values.deployment.topologySpreadConstraints }}
      topologySpreadConstraints:
        {{- toYaml . | nindent 8 }}
      {{- end }}
```

mustafaStakater:
I will look into automating changelog updates @aslafy-z

rasheedamir:
@aslafy-z maybe we should merge this PR by updating the changelog manually and don't hold this one for automatic updates of changelog?

rasheedamir:
@aslafy-z i will suggest we add tests like these for this change? https://github.com/stakater/application/tree/master/application/tests

aslafy-z:
@rasheedamir I was not sure if such a proxified value with no transformation needed a test. What do you think? Would love to get @d3adb5 's opinion too

d3adb5:
I'm with @aslafy-z on this one. Trying to think of it in terms of what we might test in a real program, there's no logic beyond accessing variables and ensuring `with` works as intended — which is language behavior — to test, so the unit test would be trivial and unnecessary at best.

d3adb5 (approved):

PR #246, opened 2023-04-02
ci/cd: simplify and refactor workflows

d3adb5:
We discussed switching from running a dry-run install on the Single Node OpenShift clusters as self-hosted runners and instead using kind or k0s.

This PR changes the CI and CD workflows to use kind for a dry-run. The CI workflow is now invoked from the CD workflow to avoid repetition, and Tilt has been removed altogether since I couldn't see why the Tilt setup was necessary.

Bringing back the Tilt files should be as easy as running `git revert` on 645165f. If we end up having to use it again, that's how we can restore them.

Additionally, linting, unit testing, and the dry run install happen in parallel now. Only the latter needs a cluster, so that's where we bring up kind. For now, since we're not testing specific cluster topologies, no cluster configuration has been added.


d3adb5:
It seems every time I read over the documentation and reference for GitHub Actions I find something has changed. I'd love to get some input from everybody on these changes. @rasheedamir @aslafy-z @slauger Can you guys leave a review?

The updated actions won't run here until merged, so I tested them on my fork:

- https://github.com/d3adb5/application/actions/runs/4589451185
- https://github.com/d3adb5/application/actions/runs/4589447205

And in the `#testing-pr-notifications` channel on Stakater's Slack community.

d3adb5:
About the failed status check: the Tilt files were removed in this PR, so `tilt ci` will definitely fail. I'm not sure how to get GitHub Actions to run the updated workflows here. :confused:

rasheedamir:
Lets do it incrementally; we keep the current workflows as is and add new ones and when the new one has all features of the existing workflow then we can remove

We want to test and validate everything i.e. route; grafana dashboard; etc.

We use tilt to setup stuff on the SNO

rasheedamir:
also we have a plan that we should add real integration tests as well i.e. we deploy one sample application and then validate all of the resources are created and are valid with tests; so, we can guarantee 100% quality of this chart

d3adb5:
> We want to test and validate everything i.e. route; grafana dashboard; etc.

I see! Does Tilt test those automatically when we run `tilt ci`? I'm pretty unfamiliar with Tilt, so it wasn't clear to me by looking at the current workflows.

aslafy-z (dismissed):

mustafaStakater:
how do we plan on making the missing crds available in kind?

rasheedamir (changes requested):
Missing CRDs in kind how do we resolve it? Also the goal is that we are testing on real cluster so, can guarantee the quality of the chart

karl-johan-grahn:
Talked about it at Open Source pruning session today:
- Review the need for this change
- Feasibility of installing all CRDs on a temporary cluster

PR #247, opened 2023-04-03
Revert "Defining default path spec in ingress if hosts.*.paths is null"

mustafaStakater:
This reverts PR https://github.com/stakater/application/pull/240.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 All notable changes to this project will be documented here.
 
 ### v2.2.1
-- Fix: Reverts, This can already be configured via `paths` [PR-243](https://github.com/stakater/application/pull/243) [PR-247](https://github.com/stakater/application/pull/247)
+- Fix: Reverts [PR-243](https://github.com/stakater/application/pull/243), This can already be configured via `paths` [PR-247](https://github.com/stakater/application/pull/247)
 
 ### v2.2.0
 - Feature: Add topologySpreadConstraints [PR-239](https://github.com/stakater/application/pull/239)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 All notable changes to this project will be documented here.
 
 ### v2.1.19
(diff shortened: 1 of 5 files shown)


d3adb5 (approved):
I personally don't see the problem with allowing use of a simpler interface for this kind of thing, though I wonder how often it would be used and whether its use should even be recommended over reproducing the classic Ingress path structure.

Nevertheless, I'm approving this one for now, but would also like to hear @rasheedamir's and @aslafy-z's opinions.

mustafaStakater:
hmmmm, It creates redundancy in a sense that we can already define the same configuration via `paths` value. 

I will be removing the else block so that users are forced to define basic `paths` value to avoid ambiguity. @d3adb5

aslafy-z:
@mustafaStakater @d3adb5 I'd prefer to only expose the standard ingress path format as values. I think we can drop `servicePort` and `serviceName` too. However, if we go that way, it becomes a breaking change, thus, we'll have to bump the major.

aslafy-z:
By "expose the standard ingress path format as values" I meant expose the whole `paths` list in the values as done in this revert. Ok to me.

aslafy-z (approved):

CHANGELOG.md
```diff
@@ -3,11 +3,14 @@
 
 All notable changes to this project will be documented here.
 
+### v2.1.14
+- Fix: Reverts [PR-243](https://github.com/stakater/application/pull/243), This can already be configured via `paths` [PR-247](https://github.com/stakater/application/pull/247)
```
aslafy-z, line 7:
```suggestion
- Fix: Reverts [PR-240](https://github.com/stakater/application/pull/240), This can already be configured via `paths` [PR-247](https://github.com/stakater/application/pull/247)
```

aslafy-z (dismissed):

d3adb5 (approved):
I feel like this could break someone's existing setup, if they're reliant on the path fallback. It's not exactly warranting of a major bump, but caution is always good.

rasheedamir:
@mustafaStakater what do you think? should we do major bump?

mustafaStakater:
minor bump looks ok to me

mustafaStakater:
we can merge this

hussnain612 (approved):

Issue #32, opened 2020-06-07
Add VPA manifest


rasheedamir:
Add VPA manifest; and I would like it to be always running in recommendation mode and document it as well

---

In this example, you create a VerticalPodAutoscaler that has an updateMode of "Off". Then you create a Deployment that has two Pods, each of which has one container. When the Pods are created, the VerticalPodAutoscaler analyzes the CPU and memory needs of the containers and records those recommendations in its status field. The VerticalPodAutoscaler does not take any action to update the resource requests for the running containers.

Here is a manifest for the VerticalPodAutoscaler:

```
apiVersion: autoscaling.k8s.io/v1
kind: VerticalPodAutoscaler
metadata:
  name: my-rec-vpa
spec:
  targetRef:
    apiVersion: "apps/v1"
    kind:       Deployment
    name:       my-rec-deployment
  updatePolicy:
    updateMode: "Off"
```

Here is a manifest for the Deployment:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-rec-deployment
spec:
  replicas: 2
  selector:
    matchLabels:
      app: my-rec-deployment
  template:
    metadata:
      labels:
        app: my-rec-deployment
    spec:
      containers:
      - name: my-rec-container
        image: nginx
```

In the manifest, you can see that there are no CPU or memory requests. You can also see that the Pods in the Deployment belong to the VerticalPodAutoscaler, because it points to the target of kind: Deployment and name: my-rec-deployment.

The output shows recommendations for CPU and memory requests:

```
...
  recommendation:
    containerRecommendations:
    - containerName: my-rec-container
      lowerBound:
        cpu: 25m
        memory: 262144k
      target:
        cpu: 25m
        memory: 262144k
      upperBound:
        cpu: 7931m
        memory: 8291500k
...
```

PR #249, opened 2023-04-03
feat: Add vertical pod autoscaler template

OsamaKhan220:
Fixes https://github.com/stakater/application/issues/32

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,6 +302,15 @@ Stakater [Forecastle](https://github.com/stakater/Forecastle) parameters
 | autoscaling.minReplicas | Sets minimum replica count when autoscaling is enabled                                                                                                                                           | `1`                                                                                                                                                   |
 | autoscaling.maxReplicas | Sets maximum replica count when autoscaling is enabled                                                                                                                                           | `10`                                                                                                                                                  |
 | autoscaling.metrics | Configuration for hpa metrics, set when autoscaling is enabled                                                                                                                                   | `{}`                                                                                                                                                  |
+### VPA - Vertical Pod Autoscaler Paramaters
+
+| Name                     | Description                                                                                  | Value           |
+| ------------------------ | -------------------------------------------------------------------------------------------- | --------------- |
+| vpa.enabled              | Enable vertical pod autoscaler                                                               | `false`         |
+| vpa.additionalLabels     | Labels for vertical pod autoscaler                                                           | `{}`            |
+| vpa.annotations          | Annotations for vertical pod autoscaler                                                      | `{}`            |
+| vpa.containerPolicies    | container policies for individual containers.                                                | `[]`            |
+
 
 ### EndpointMonitor Paramaters
 
--- a/README.md
+++ b/README.md
@@ -302,7 +302,8 @@ Stakater [Forecastle](https://github.com/stakater/Forecastle) parameters
 | autoscaling.minReplicas | Sets minimum replica count when autoscaling is enabled                                                                                                                                           | `1`                                                                                                                                                   |
 | autoscaling.maxReplicas | Sets maximum replica count when autoscaling is enabled                                                                                                                                           | `10`                                                                                                                                                  |
 | autoscaling.metrics | Configuration for hpa metrics, set when autoscaling is enabled                                                                                                                                   | `{}`                                                                                                                                                  |
-### VPA - Vertical Pod Autoscaler Paramaters
+
+### VPA - Vertical Pod Autoscaler Parameters
 
 | Name                     | Description                                                                                  | Value           |
 | ------------------------ | -------------------------------------------------------------------------------------------- | --------------- |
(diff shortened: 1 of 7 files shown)


application/templates/vpa.yaml
```diff
@@ -0,0 +1,24 @@
+{{- if and (.Values.vpa).enabled (.Capabilities.APIVersions.Has "autoscaling.k8s.io/v1") -}}
```
mustafaStakater, line 1: we should specify `kind` in capabilities check because `autoscaling.k8s.io` might exist but vpa crd might not

d3adb5, line 1: This could potentially be misleading: if a user sets `vpa.enabled` to `true`, it could still be that there is no `VerticalPodAutoscaler` resource on their cluster, making it so Helm simply doesn't render the template and the user might end up confused as to what is wrong.

I suggest adding a check instead:

```suggestion
{{- if .Values.vpa.enabled -}}

{{- if not (.Capabilities.APIVersions.Has "autoscaling.k8s.io/v1/VerticalPodAutoscaler) }}
  {{- fail "There is no VerticalPodAutoscaler resource definition in the target cluster!" }}
{{- end }}

```

Additionally, a unit test should be added to check that such a validation is actually performed.

d3adb5 (dismissed):

mustafaStakater:
Add `helm uninstall openshift-vertical-pod-autoscaler -n openshift-vertical-pod-autoscaler` command in Tiltfile-delete.

README.md
```diff
@@ -302,6 +302,15 @@ Stakater [Forecastle](https://github.com/stakater/Forecastle) parameters
 | autoscaling.minReplicas | Sets minimum replica count when autoscaling is enabled                                                                                                                                           | `1`                                                                                                                                                   |
 | autoscaling.maxReplicas | Sets maximum replica count when autoscaling is enabled                                                                                                                                           | `10`                                                                                                                                                  |
 | autoscaling.metrics | Configuration for hpa metrics, set when autoscaling is enabled                                                                                                                                   | `{}`                                                                                                                                                  |
+### VPA - Vertical Pod Autoscaler Paramaters
```
aslafy-z, line 305:
```suggestion

### VPA - Vertical Pod Autoscaler Parameters
```

application/templates/vpa.yaml
```diff
@@ -0,0 +1,27 @@
...
+    {{- toYaml .Values.vpa.annotations | nindent 4 }}
+  {{- end }}
+spec:
+  targetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: {{ template "application.name" . }}
+  resourcePolicy:
+    containerPolicies:
+      {{- toYaml .Values.vpa.containerPolicies | nindent 6 }}
```
aslafy-z, line 26: Shouldn't this fail if no `containerPolicies` are configured?

Tiltfile
```diff
@@ -40,5 +40,20 @@ local_resource(
...
+    )
+
+# Wait until VPA CRD becomes available
+local_resource(
+    'wait-for-crds', 
+    cmd='timeout 300s bash -c "until kubectl wait --for condition=Established crd/verticalpodautoscalers.autoscaling.k8s.io; do sleep 10; done"',
+    resource_deps=[
+        'openshift-vertical-pod-autoscaler'
+    ])
 # Install cert-manager
```
aslafy-z, line 58:
```suggestion

# Install cert-manager
```

aslafy-z (dismissed):

Tiltfile-delete
```diff
@@ -8,6 +8,10 @@ def delete_external_secrets_operator():
     print('Deleting external-secrets operator')
     local('helm uninstall external-secrets-operator -n external-secrets-operator')
 
+def delete_openshift_vertical_pod_autoscaler():
+    print('Deleting external-secrets operator')
```
d3adb5, line 12: Is this correct?

```suggestion
    print('Uninstalling openshift-vertical-pod-autoscaler')
```

Tiltfile
```diff
@@ -48,5 +48,13 @@ local_resource(
     cmd='helm install openshift-vertical-pod-autoscaler -n openshift-vertical-pod-autoscaler oci://ghcr.io/stakater/charts/openshift-vertical-pod-autoscaler'
     )
 
+# Wait until VPA CRD becomes available
+local_resource(
+    'wait-for-crds', 
+    cmd='timeout 300s bash -c "until kubectl wait --for condition=Established crd/verticalpodautoscalers.autoscaling.k8s.io; do sleep 10; done"',
+    resource_deps=[
+        'openshift-vertical-pod-autoscaler'
+    ])
```
d3adb5: Would it make sense to kind of merge these two commands together? Just a small suggestion, in case it makes semantic sense. I'm not well versed in Tilt, so this might be silly.

d3adb5 (changes requested):
I added a prompt that could become a suggestion, and apart from what I think is an erroneous string, this looks good to me.

Tiltfile
rasheedamir: @d3adb5 can you plz create separate ticket for this so, we can merge this one?

d3adb5: I don't see a reason to create an issue for what I think is a style choice, so I'll just resolve the thread for now and leave my approval. :+1:

d3adb5 (dismissed):
LGTM, minus the currently unavoidable merge conflict. :+1:

aslafy-z (approved):

PR #253, opened 2023-05-13
Manage Release name / selector labels / add label part-of / component

c-dst:
Hi, 

I used the stakater application charts for some deployement. 
I had see a limitation in some case. When I wanted to deploy the same application multiple time in the name namespace, because the charts didn't use the Helm Release.Name.

So I updated the _helper.tpl to generate an application fullname which use the Helm Release.Name and the applicationName. 

So the name of the deployed objects are {{ Release.Name }}-{{ applicationName }}

It's now possible to deploy multiple instance of the same value file in the same namespace. 

To to that 
I refactor labels and selectorLabels with 
 - app.kubernetes.io/instance:
 - app.kubernetes.io/component:
 - app.kubernetes.io/part-of:

Updated templates and value.yaml remove all trailing spaces.

### DIFF
--- a/application/templates/_helpers.tpl
+++ b/application/templates/_helpers.tpl
@@ -8,6 +8,19 @@ Define the name of the chart/application.
 {{- end -}}
 
 {{/*
+Define a fully qualified app name with Release.Name and chart/application name
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "application.fullname" -}}
+{{- $name := default .Chart.Name .Values.applicationName -}}
+{{- if contains $name .Release.Name -}}
+{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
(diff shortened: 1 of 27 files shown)


d3adb5:
I haven't had the time to look over all the changes yet, but I'll leave a few thoughts here:

1. Please avoid committing whitespace changes alongside meaningful ones. It messes with blame information and makes it more difficult to pinpoint what the origin of each line is, later down the line. I suggest removing the trailing whitespace changes from your PR, or moving them to a separate commit.
2. Since these changes will result in resources with different names, as well as references to `Secret`'s with different names, **we might need to do a major release bump** if we accept it, as current users might find a new release of the chart to cause some downtime.
3. Some of these changes are unnecessary and only produce noise (e.g. the main container's name in the `Deployment` template). Please review whether using the fully qualified name is necessary to make the distinction between two releases in the same namespace so that the minimum amount of changes is made.

PR #255, opened 2023-07-05
fix: allow setting arbitrary resource constraints

aslafy-z:
This PR allows to request additional or non standard resources.
Also allows unsettings default resources from values using `null`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,13 +3,11 @@
 
 All notable changes to this project will be documented here.
 
-
 ### v2.1.15
 - fix: allow setting arbitrary resource constraints [PR-255](https://github.com/stakater/application/pull/255)
 
 ### v2.1.14
 - feat: allow overriding clusterIP [PR-258](https://github.com/stakater/application/pull/258)
-
 
 ### v2.1.13
 - Fix: make deployment.image.tag and job.image.tag optional [PR-234](https://github.com/stakater/application/pull/234)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 
 All notable changes to this project will be documented here.
 
+### v2.1.14
+- fix: allow unseting resources [PR-255](https://github.com/stakater/application/pull/255)
+
 ### v2.1.13
 - Fix: make deployment.image.tag and job.image.tag optional [PR-234](https://github.com/stakater/application/pull/234)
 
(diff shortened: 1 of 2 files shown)


application/templates/deployment.yaml
```diff
@@ -246,13 +246,10 @@ spec:
         {{- end }}
         {{- end }}
       {{- end }}
+        {{- if .Values.deployment.resources }}
```
d3adb5, line 249: We can turn this into a `with` block instead! (`.`) would be used within.

```suggestion
        {{- with .Values.deployment.resources }}
```

application/templates/deployment.yaml
```diff
@@ -246,13 +246,10 @@ spec:
...
         resources:
-          limits:
-            memory: {{ .Values.deployment.resources.limits.memory }}
-            cpu: {{ .Values.deployment.resources.limits.cpu }}
-          requests:
-            memory: {{ .Values.deployment.resources.requests.memory }}
-            cpu: {{ .Values.deployment.resources.requests.cpu }}
+{{ toYaml .Values.deployment.resources | indent 10 }}
```
d3adb5, line 251: Accompanying the `with` suggestion.

```suggestion
          {{- toYaml . | nindent 10 }}
```

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 
 All notable changes to this project will be documented here.
 
+### v2.1.14
+- fix: allow unseting resources [PR-255](https://github.com/stakater/application/pull/255)
```
d3adb5, line 7: Better than allowing no resources to be set, this allows arbitrary resources to be configured. This way, if new resource paremeters are added, it'll be possible to set them this way.

The `ephemeralStorage` resource parameter has existed for a while, but it's not possible to set it on the Deployment resource currently. This would fix that. Maybe the change message could be updated to display this broad application to the resources value?

aslafy-z, line 7:
```suggestion
- fix: allow setting arbitrary resources [PR-255](https://github.com/stakater/application/pull/255)
```

aslafy-z: Let me know if it fits you!

d3adb5: Maybe "arbitrary resource constraints", for added clarity?

d3adb5 (dismissed):
Left a comment in an earlier thread. Otherwise, LGTM!

CHANGELOG.md
aslafy-z: Done 👍

d3adb5 (approved):

Issue #245, opened 2023-04-01
Limit deployment history


rasheedamir:
Pods can be slow when creating, updating, or deleting because old objects are still tracked in the cluster. You can reduce the `revisionHistoryLimit` of [deployments](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#clean-up-policy) to cleanup older ReplicaSets which will lower to total amount of objects tracked by the Kubernetes Controller Manager. The default history limit for Deployments in 10.

Our recommended value should be 2 and not more

PR #257, opened 2023-07-05
feat: allow overriding revisionHistoryLimit

aslafy-z:


application/values.yaml
```diff
@@ -187,6 +187,9 @@ deployment:
     #       values:
     #       - ssd
 
+  # Number of ReplicaSet versions to retain
+  revisionHistoryLimit: 2
```
d3adb5, line 191: I understand that #245 stipulates a default value of 2 here for performance gains in avoiding keeping old `ReplicaSets` around. However, one of the reasons to use Kubernetes is to have an easier time rolling changes back. Could this value affect the possibility of rollbacks through Helm, or, if any brave soul is mixing it into production, `kubectl rollback`?

This is of course not enough to warrant a _major_ version bump, but in my opinion a warning about this new default value and its repercussions might be warranted in the project's README.

d3adb5 (changes requested):

application/values.yaml
aslafy-z: What do you think about adding this warning in the Changelog instead?

d3adb5 (approved):
LGTM!

rasheedamir:
@aslafy-z plz update the changelog then lets get this merged as well

rasheedamir (approved):

PR #271, opened 2023-09-22
fix: allow multiple cronjobs

aslafy-z:
The cronjob template misses a `---` new document separation. This means all cronjobs are scrambled together and only a frankenstein version of all of them is emitted. 
This PR also does some lint:
- move the new document separator to the beginning of templates
- add empty newline at the end of files
- remove empty newlines in the middle of files
- move helpers to _helpers file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 
 All notable changes to this project will be documented here.
 
+### v2.2.6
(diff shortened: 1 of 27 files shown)


d3adb5 (approved):
I would normally be wary of changes to whitespace, as they unnecessarily mess with blame information, but in this case they're only present at the end of each file and I doubt we'll run into any issues as a result.

LGTM! :+1:

PR #282, opened 2023-10-12
allow stringData on secrets

emmanuelm41:


rasheedamir:
@d3adb5 @aslafy-z please review it

application/templates/secret.yaml
```diff
@@ -15,9 +15,21 @@ metadata:
...
+{{- if .data }}
 data:
 {{- range $key, $value := .data }}
   {{ $key }}: {{ $value | b64enc }}
 {{- end }}
+{{- else if .stringData }}
+stringData:
+{{- range $key, $value := .stringData }}
+  {{ $key }}: {{ $value }}
+{{- end }}
```
d3adb5, line 27: `stringData` isn't mutually exclusive with `data` for Secrets. We can **and should** add it alongside `data` and the API will encode the contents to base 64.

application/templates/secret.yaml
```diff
@@ -15,9 +15,21 @@ metadata:
...
+{{- else if .encodedData }}
+data:
+{{- range $key, $value := .encodedData }}
+  {{ $key }}: {{ $value }}
+{{- end }}
+{{- end }}
```
d3adb5, line 33: This feels like a similar case, where there is no problem with providing this functionality (adding base64-encoded content) alongside existing ones (plain text).

Could you come up with a way to provide Base64-encoded strings alongside plain text to create a `Secret` through this chart?

application/templates/secret.yaml
d3adb5, line 27: I feel like these are now doing the exact same thing. While I'm all for options and choice, is there a real benefit to maintaining the status quo (encoding through Helm) when the Kubernetes API provides the write-only `stringData`? I also posit that perhaps `stringData` itself isn't necessary since it's just a way to delegate what we're already doing to the API server.

application/templates/secret.yaml
d3adb5, line 1: Please add unit tests when you have branching template code to ensure you're getting the expected behavior.

d3adb5 (changes requested):
Thank you for your contribution, @emmanuelm41! I added a few comments below.

application/templates/secret.yaml
emmanuelm41: Done!

application/templates/secret.yaml
emmanuelm41: Done!

application/templates/secret.yaml
emmanuelm41: Added unit test :D

application/templates/secret.yaml
emmanuelm41: Let me explain our requirement. We use 1pass vault plugin for Kubernetes, and the actual value is a placeholder that will be replaced by the plugin (it is an operator). Therefore, we need to set a string value, and leave it without being encoded at all, as the real value will be replaced by the plugin.

emmanuelm41:
@d3adb5 I just applied a few changes, and add a couple of unit tests. Could you please give me your feedback now?

application/templates/secret.yaml
d3adb5: Interesting! I'm not opposed to the change, but from what you're telling me I think you'd need an admission hook, no? When writing a `Secret` with `stringData` in it, the API server encodes the plain text into base 64 and then makes it a regular `data` key. According to the documentation, `stringData` is never output by the API.

Does this fit your scenario? I'm not sure if you could see `stringData` in an admission hook, but I bet that's what you might have to work with.

application/values-test.yaml
d3adb5, line 1: It looks like all the changes in this file have to do with trailing whitespace. Could you revert that in the commits submitted as part of this pull request so they don't mess with git-blame information?

If you don't want to amend past commits, you can revert these in a new commit and squash everything before we merge.

application/templates/secret.yaml
```diff
@@ -15,9 +15,24 @@ metadata:
...
+{{- if .stringData }}
+stringData:
+{{- range $key, $value := .stringData }}
+  {{ $key }}: {{ $value }}
+{{- end }}
+{{- end }}
+{{- end }}
+{{- end }}
```
d3adb5, line 37: It looks like this first `if` is closed at the wrong place: the way these conditionals are set up, data or string data are only included in this Secret if `.data` or `.encodedData` are truth-y. That said, I suggest avoiding conditionals altogether:

```suggestion
data:
  {{- range $key, $value := .data }}
  {{ $key }}: {{ $value | b64enc }}
  {{- end }}
  {{- range $key, $value := .encodedData }}
  {{ $key }}: {{ $value }}
  {{- end }}
stringData:
  {{- toYaml .stringData | nindent 2 }}
```

This way, `data` will turn `null` due to a completely empty value in the rendered template if neither `.data` nor `.encodedData` are populated, and `stringData` will be an explicit `null` if `.stringData` is itself a null value.

What do you think? I think the unit test you wrote will still work.

d3adb5 (changes requested):
I left a suggestion and a request for change.

d3adb5:
Hi @emmanuelm41, have you had the chance to review my comments? If anything's unclear, please let me know so I can help clarify what I said.

application/values-test.yaml
emmanuelm41: Nice change. Smart one! Great!

application/templates/secret.yaml
emmanuelm41: Nice change. Smart one! Great!

application/templates/secret.yaml
emmanuelm41: i am not sure what you actually mean unfortunately. Could you explain me more? We are currently using the chart with this new feature on our k8s cluster from a local registry. I know it works the way we need because of it.

emmanuelm41:
> Hi @emmanuelm41, have you had the chance to review my comments? If anything's unclear, please let me know so I can help clarify what I said.

Sorry for my late response. I just applied your suggestion. Please, let's review it and see if we can merge. Thanks!

application/templates/secret.yaml
d3adb5: My assumption when you mentioned the 1pass vault plugin behaved as an operator was that it would monitor Kubernetes resources as they were ingested into etcd and read from the API server. The documentation is clear that `stringData` for secrets are never obtained from the API server and exist solely as a write-only convenience method.

To be able to capture the field, therefore, one would have to create an admission hook, which runs before resources are sent to etcd to be read as internal state through the API server. However, since you mentioned your setup already works, I'll assume there's something I'm not seeing and the plugin works its way into the K8s API rather than behave as the kind of operator I'm used to!

For now I'm resolving this thread. You can disregard what I said!

application/values-test.yaml
d3adb5: @emmanuelm41 This file still has whitespace-only changes. :frowning_face:

d3adb5, line 1: I still think these whitespace-only changes should be reverted before merging.

d3adb5 (approved):
Left a comment about whitespace-only changes, but otherwise this looks good to me! Thank you, @emmanuelm41!

application/values-test.yaml
emmanuelm41: Fixed!

d3adb5 (dismissed):
LGTM!

d3adb5:
@rasheedamir, @aslafy-z, can anybody else take a quick look and figure out if we can merge this?

karl-johan-grahn:
@emmanuelm41 can you please resolve the merge conflicts?

emmanuelm41:
@karl-johan-grahn Done!

application/values-test.yaml
mustafaStakater, line 1: Need to update values.yaml file, add comments elaborating data, encodedData, stringData

mustafaStakater:
Please update the README.md and CHANGELOG

application/Chart.yaml
```diff
@@ -6,7 +6,7 @@ type: application
 
 # Helm chart Version
 
-version: 2.2.13
+version: 2.3.0
```
mustafaStakater, line 9: add `#minor` in commit message to bump the minor version instead of setting version to `2.3.0`

emmanuelm41: This is done!

application/values-test.yaml
emmanuelm41: done!

emmanuelm41:
> @emmanuelm41 can you please look into resolving the comments and merge conflicts?

Everything should be covered now!

application/values.yaml
d3adb5, line 1: This file has meaningful changes, but also involves the removal of trailing whitespace. As mentioned before, this can mess with blame information, if we use it in this project to track down which commits introduced specific changes.

d3adb5 (dismissed):
This looks good to me. I left a comment about whitespace changes in the chart's values file, but it could be considered nitpicking a little.

mustafaStakater (commented):
Looks good @emmanuelm41 , is it possible to resolve @d3adb5 last comment, and we can proceed with merging this

application/values.yaml
emmanuelm41: Done

emmanuelm41:
> Looks good @emmanuelm41 , is it possible to resolve @d3adb5 last comment, and we can proceed with merging this

Resolved!

mustafaStakater (approved):
Thanks for the contribution – your efforts are appreciated :+1:

PR #290, opened 2023-12-05
Fix lint

ElfoLiNk:
```
    [ERROR] templates/ingress.yaml: unable to parse YAML: invalid Yaml document separator: apiVersion: networking.k8s.io/v1
```


aslafy-z:
I updated your change to make future changes less error prone. 
For some reason, stakater runners are not running jobs. I'll loop with their team and make sure this PR gets merged.
Thank you for the report!

PR #292, opened 2023-12-27
Allowing HPA behavior to be set

ben851:
Adding the ability to set autoscaling behavior as per https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#configurable-scaling-behavior


rasheedamir:
@aslafy-z please review and merge

aslafy-z (approved):

PR #317, opened 2024-06-12
add schedule

AsfaMumtaz:


application/values-test.yaml
```diff
@@ -1307,7 +1307,7 @@ cronJob:
             cpu: 1
 
 # Take backup of application namespace
-backup:
+schedule:
```
d3adb5, line 1310: Instead of `schedule` we can be a little more explicit about what this is for in the value's own name.

```suggestion
backupSchedule:
```

What do you think, @AsfaMumtaz?

application/values.yaml
```diff
@@ -906,8 +906,8 @@ grafanaDashboard:
     #       app: test-2
 
 ##########################################################
-# Backup object for creating back using OADP/Velero
+# Backup schedule object for creating back using OADP/Velero
 ##########################################################
-backup:
+schedule:
```
d3adb5, line 911: Same suggestion as in the other thread I opened.

d3adb5 (changes requested):
I think we're missing updates to the `README.md`  with this change. It still mentions `backup.*` in the values tables.

d3adb5:
I forgot to mention one thing: this changes the interface in a way that is not backwards compatible, so it would require a major version bump. How about instead we detect whether `.Values.backup` is filled and use it if present as long as `.Values.schedule` (or `.Values.backupSchedule`) is empty? That would allow us to deprecate it, issue a warning in the NOTES.txt template, and in the next major version upgrade do away with the old option completely.

PR #320, opened 2024-06-28
Make cronJob work without a service account

TomyLobo:
When using a cronJob without configuring a serviceAccount, I get the following error under "Events" in "kubectl describe":
```
  Warning  FailedCreate  5m33s (x534 over 2d4h)  job-controller  Error creating: pods "myapp-weekly-dst-28656480-" is forbidden: error looking up service account myapp/myapp: serviceaccount "myapp" not found
```
This PR fixes that by evaluating `rbac.serviceAccount.enabled` instead of `rbac.enabled`, bringing it into alignment with deployment.yaml.

### DIFF
--- a/application/templates/cronjob.yaml
+++ b/application/templates/cronjob.yaml
@@ -54,7 +54,7 @@ spec:
           annotations: {{ toYaml . | nindent 12 }}
           {{- end }}
         spec:
-          {{- if $.Values.rbac.enabled }}
+          {{- if $.Values.rbac.serviceAccount.enabled }}
           {{- if $.Values.rbac.serviceAccount.name }}
           serviceAccountName: {{ $.Values.rbac.serviceAccount.name }}
             {{- else }}


TomyLobo:
There's a check failing that has nothing to do with my changes.
Not sure what to do.
Help? :)

application/templates/cronjob.yaml
```diff
@@ -54,7 +54,7 @@ spec:
           annotations: {{ toYaml . | nindent 12 }}
           {{- end }}
         spec:
-          {{- if $.Values.rbac.enabled }}
+          {{- if $.Values.rbac.serviceAccount.enabled }}
```
d3adb5, line 57: Shouldn't we still keep the `if $.Values.rbac.enabled` line? If RBAC for this chart has been completely disabled, it shouldn't in theory matter if `rbac.serviceAccount.enabled` is true, is my thinking.

d3adb5 (commented):
It looks like the problem with the CI workflow is infrastructure (configuration) related. I don't have access to the runners these jobs are running on, but given the error message (missing versions of dynamic glibc) I assume the runner is perhaps on CentOS 7 while the new `oc` binary was built with RHEL9 in mind --- hence the newer versions of glibc required. Can you guys take a look, @rasheedamir? This is affecting ALL builds. :sweat_smile:

d3adb5:
FYI I opened an issue for this: #321. For now let's wait for that to be fixed so we don't merge anything without all the automated tests passing, @TomyLobo.

application/templates/cronjob.yaml
TomyLobo: Same goes for the deployment, then.
I didn't check if there are other parts as well, which would be equally affected.
I'm just building on your reasoning for the deployment :)

aslafy-z:
@TomyLobo I just rebased the PR (pull with `git pull --rebase` next time). Tests are running!

aslafy-z:
Superseeded by https://github.com/stakater/application/pull/363, which will soon be merged.

PR #335, opened 2024-07-16
feat: generated Helm documentation

### LINKED ISSUE #242: Automate helm documentation update
There are some tools out there to automatically generate helm chart documentation from the values and comments. It would ensure we don't miss any entries and enforce comments in the value file.

aslafy-z:
Preview at https://github.com/stakater/application/blob/feat/generated-readme/README.md

- Generate README.md with https://github.com/norwoodj/helm-docs
- Add pre-commit hook and GitHub Action run to validate README
- Document the whole values file
- Move some defaults from templates to values
- Set `certificate.issuerRef.group` to `cert-manager.io`

Thanks to https://github.com/norwoodj/helm-docs/pull/260, we will soon be able to render a bit better some of the default values.

Fixes #242

### DIFF
--- a/.github/workflows/pull_request.yaml
+++ b/.github/workflows/pull_request.yaml
@@ -73,3 +73,19 @@ jobs:
         with:
           charts: application
           unittest-version: v0.5.x
+
+  check-helm-docs:
+    name: Check Helm Docs
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{github.event.pull_request.head.sha}}
+      - name: Run helm-docs and check diff
+        # TODO: Move to upstream once https://github.com/losisin/helm-docs-github-action/pull/270 is merged
+        uses: losisin/helm-docs-github-action@0f2a7b456e9e4393faa24d4d2e1636bdb0a5b9b9
+        with:
+          output-file: ./README.md
+          template-files: ./README.md.gotmpl
+          sort-values-order: file
+          fail-on-diff: true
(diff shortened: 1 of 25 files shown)


rasheedamir:
@karl-johan-grahn @d3adb5 can you guys review it?

karl-johan-grahn (dismissed):

application/templates/alertmanagerconfig.yaml
```diff
@@ -6,15 +6,15 @@ metadata:
   name: {{ template "application.name" . }}
   namespace: {{ include "application.namespace" . }}
   labels:
-  {{- include "application.labels" . | nindent 4 }}
+  {{- include "application.labels" $ | nindent 4 }}
```
SheryarButt, line 9: Can you explain what $ does here that wasn't already covered by `.` ?

aslafy-z: This change and some other similar ones can be reverted. I mass replaced include statements. Using $ there ensures that even if they are wrapped in a range in the future, they get th le right context.

.github/workflows/pull_request.yaml
```diff
@@ -73,3 +73,17 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{github.event.pull_request.head.sha}}
+      - name: Run helm-docs and check diff
+        uses: losisin/helm-docs-github-action@v1
+        with:
+          output-file: README.md
+          template-files: README.md.gotmpl
+          fail-on-diff: true
```
karl-johan-grahn, line 89: Why should it fail on diff?

aslafy-z: It makes the status check fail if the values changed but the documentation was not updated in the PR

karl-johan-grahn: ah ok nice

rasheedamir:
looks good to me! great work @aslafy-z

PR #352, opened 2024-10-01
feat(deployment): allow setting replicas to zero #minor

ElfoLiNk:


application/templates/deployment.yaml
```diff
@@ -20,7 +20,7 @@ metadata:
   name: {{ template "application.name" . }}
   namespace: {{ include "application.namespace" . }}
 spec:
-{{- if .Values.deployment.replicas }}
+{{- if not (quote .Values.deployment.replicas | empty) }}
```
aslafy-z, line 23:
```suggestion
{{- if not (eq .Values.deployment.replicas nil) }}
```
I would rather check vs nil instead of empty.

aslafy-z:
This was released with https://github.com/stakater/application/releases/tag/v5.1.0. Thank you for your contribution.

PR #354, opened 2024-10-28
feat(cronjob): Corrected indentation of securityContext container params #minor

epm-marcus:
Corrected the indentation level of securityContext to instead apply it to each container within the jobs.


karl-johan-grahn (approved):

aslafy-z:
Why was this PR merged?
It is a breaking change. 
Job can have a pod level or container level security context. 
I think it should be reverted and a new PR should be emitted to add containerSecurityContext.

PR #355, opened 2024-11-06
tenant namespaces can be added in includeNamespace while the default backup namespace would be the -system namespace

TehreemNisa:
…


John2020-cyber (approved):

aslafy-z:
This update is NOT good! No documentation was added for the field, the naming is wrong, it only work with a single namespace.
How can this be merged in this state?

PR #360, opened 2024-11-13
fix: Remove default values for rollingUpdate

arielly-parussulo:
RollingUpdate is an optional property only used when we set the strategy type as "RollingUpdate". If we set the type as "Recreate", this property breaks the deployments wit the following message:

Deployment.apps "<app>" is invalid: spec.strategy.rollingUpdate: Forbidden: may not be specified when strategy `type` is 'Recreate'

It gets more problematic when we use stakater/apps Helm Chart as a subchart for our applications. Helm has a known bug that doesn't allow override values from the subchart. So, if we need to deploy an application with the Recreate strategy, it is not possible as the deployment will always break with the error mentioned above.

The simpler solution is to comment the "rollingUpdate" property to prevent the Recreate deployments to break when using a subchart.

### DIFF
--- a/application/values.yaml
+++ b/application/values.yaml
@@ -92,13 +92,13 @@ deployment:
     # -- (string) Type of deployment strategy.
     # @section -- Deployment Parameters
     type: RollingUpdate
-    rollingUpdate:
+    # rollingUpdate:
       # -- (string) Max unavailable pods during update.
       # @section -- Deployment Parameters
-      maxUnavailable: 25%
+      # maxUnavailable: 25%
       # -- (string) Max surge pods during update.
       # @section -- Deployment Parameters
-      maxSurge: 25%
+      # maxSurge: 25%
   # -- (bool) Reload deployment if attached Secret/ConfigMap changes.
   # @section -- Deployment Parameters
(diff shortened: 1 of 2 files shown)


application/values.yaml
```diff
@@ -92,13 +92,13 @@ deployment:
...
-    rollingUpdate:
+    # rollingUpdate:
       # -- (string) Max unavailable pods during update.
       # @section -- Deployment Parameters
-      maxUnavailable: 25%
+      # maxUnavailable: 25%
       # -- (string) Max surge pods during update.
       # @section -- Deployment Parameters
-      maxSurge: 25%
+      # maxSurge: 25%
```
aslafy-z, line 101: Can you please remove the documentation comments and pack the example values like in the initContainers sections below? Then, build the docs with `make build-docs` and then commit the resulting README. Thank you!

aslafy-z (approved):

rasheedamir:
@aslafy-z can we get it merged?

rasheedamir (approved):

rasheedamir:
@aslafy-z you can merge it

PR #363, opened 2024-11-15
feat!: harmonize service account binding

aslafy-z:
Closes #320
Closes #361

## Summary

- Rename `rbac.serviceAccount.enabled` to `rbac.serviceAccount.create`
- Change `deployment.automountServiceAccountToken` default from `true` to `false`
- Fix inconsistencies in serviceAccount binding across CronJob, Job, and Deployment templates by introducing a new `application.serviceAccountName` helper
- Restore per-resource `automountServiceAccountToken` toggles (defaults to `false`)
- Document missing Role and RoleBinding `rbac.annotations` and `rbac.additionalLabels` fields

## Breaking change

**1. `rbac.serviceAccount.enabled` is renamed to `rbac.serviceAccount.create`.**

If you currently have:

```yaml
rbac:
  serviceAccount:
    enabled: true
```

You must migrate to:

```yaml
rbac:
  serviceAccount:
    create: true
```

**2. `automountServiceAccountToken` now defaults to `false` everywhere.**

Previously, `deployment.automountServiceAccountToken` defaulted to `true`. It now defaults to `false` for Deployments, CronJobs, and Jobs.

If your workloads require a service account token (for example to call the Kubernetes API), you must explicitly opt in:

```yaml
deployment:
  automountServiceAccountToken: true
```

For CronJobs and Jobs, the toggle is per job entry:

```yaml
cronJob:
  jobs:
    example:
      automountServiceAccountToken: true
```

**3. ServiceAccount name resolution changed.**

The new `application.serviceAccountName` helper follows this logic:

| `rbac.serviceAccount.create` | `rbac.serviceAccount.name` | Result |
|-|-|-|
| `true` | `""` | `<release-name>` (from `application.name`) |
| `true` | `"foo"` | `"foo"` |
| `false` | `""` | `"default"` |
| `false` | `"bar"` | `"bar"` |

Previously, CronJob and Job templates did not consistently follow this resolution.

## Test plan

- [x] Helm unit tests updated for all three breaking changes
- [ ] Verify downstream charts that set `rbac.serviceAccount.enabled` are migrated
- [ ] Verify workloads that need API access explicitly set `automountServiceAccountToken: true`

### DIFF
--- a/application/templates/_helpers.tpl
+++ b/application/templates/_helpers.tpl
@@ -94,6 +94,21 @@ reference:
 {{- end }}
 
 {{/*
+Get the name of the service account to use.
+If the service account is set to be created, return the service account name or a default name.
+If the service account is not set to be created and a name is provided, return the provided name;
+otherwise, return the default namespace service account.
+*/}}
+{{- define "application.serviceAccountName" }}
+  {{- $saName := .Values.rbac.serviceAccount.name }}
+  {{- if .Values.rbac.serviceAccount.create }}
+    {{- empty $saName | ternary (include "application.name" .) (quote $saName) }}
+  {{- else }}
+    {{- empty $saName | ternary "default" (quote $saName) }}
+  {{- end }}
+{{- end }}
+
+{{/*
 Renders httpRoute rules with proper integer type for port fields.
 Usage:
 {{ include "application.httpRoute.rules" . }}
--- a/application/templates/cronjob.yaml
+++ b/application/templates/cronjob.yaml
@@ -62,14 +62,8 @@ spec:
             {{- toYaml . | nindent 12 }}
           {{- end }}
         spec:
-          automountServiceAccountToken: {{ kindIs "invalid" $job.automountServiceAccountToken | ternary true $job.automountServiceAccountToken }}
-          {{- if $.Values.rbac.enabled }}
-          {{- if $.Values.rbac.serviceAccount.name }}
-          serviceAccountName: {{ $.Values.rbac.serviceAccount.name }}
-            {{- else }}
-          serviceAccountName: {{ template "application.name" $ }}
-          {{- end }}
-          {{- end }}
+          automountServiceAccountToken: {{ $job.automountServiceAccountToken | default false }}
+          serviceAccountName: {{ include "application.serviceAccountName" $ }}
           {{- with $job.initContainers }}
           initContainers:
           {{- range $key, $value := . }}
--- a/application/templates/deployment.yaml
+++ b/application/templates/deployment.yaml
@@ -75,6 +75,8 @@ spec:
           ]
 {{- end }}
     spec:
+      automountServiceAccountToken: {{ .Values.deployment.automountServiceAccountToken }}
+      serviceAccountName: {{ include "application.serviceAccountName" $ }}
       enableServiceLinks: {{ .Values.deployment.enableServiceLinks }}
       {{- with .Values.deployment.hostAliases }}
       hostAliases: {{- toYaml . | nindent 6 }}
@@ -327,14 +329,6 @@ spec:
         {{- end }}
         {{- end }}
       {{- end }}
-      {{- if .Values.rbac.serviceAccount.enabled }}
-      {{- if .Values.rbac.serviceAccount.name }}
-      serviceAccountName: {{ .Values.rbac.serviceAccount.name }}
-      {{- else }}
-      serviceAccountName: {{ template "application.name" $ }}
-      {{- end }}
-      {{- end }}
-      automountServiceAccountToken: {{ .Values.deployment.automountServiceAccountToken }}
       {{- if .Values.deployment.hostNetwork }}
       hostNetwork: {{ .Values.deployment.hostNetwork }}
       {{- end }}
--- a/application/templates/job.yaml
+++ b/application/templates/job.yaml
@@ -42,14 +42,8 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
     spec:
-      automountServiceAccountToken: {{ kindIs "invalid" $job.automountServiceAccountToken | ternary true $job.automountServiceAccountToken }}
-      {{- if $.Values.rbac.enabled }}
-      {{- if $.Values.rbac.serviceAccount.name }}
-      serviceAccountName: {{ $.Values.rbac.serviceAccount.name }}
-        {{- else }}
-      serviceAccountName: {{ template "application.name" $ }}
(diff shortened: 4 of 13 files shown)


application/templates/_helpers.tpl
```diff
@@ -68,3 +68,18 @@ reference:
...
+
+{{- define "application.service-account-name" }}
+{{- if .Values.rbac.enabled }}
+  {{- if and .Values.rbac.serviceAccount.enabled .Values.rbac.existingServiceAccountName }}
+    {{- fail "Conflict: 'rbac.existingServiceAccountName' is set, but a new service account is being created. Please disable 'rbac.serviceAccount.enabled' or unset 'rbac.existingServiceAccountName'." }}
+  {{- end }}
+  {{- if .Values.rbac.serviceAccount.enabled }}
+    {{- default (include "application.name" .) .Values.rbac.serviceAccount.name }}
+  {{- else }}
+    {{- default "null" .Values.rbac.existingServiceAccountName }}
```
d3adb5, line 80: Edge case, may or may not be something to take into consideration: service accounts named `null`. Also, we're assuming the result of this partial template to be interpreted as YAML rather than a raw string, that may create confusion.

```suggestion
    {{- $saName := .Values.rbac.existingServiceAccountName }}
    {{- empty $saName | ternary (quote $saName) "null" }}
```

application/templates/_helpers.tpl
```diff
@@ -68,3 +68,18 @@ reference:
   kind: Route
   name: {{ include "application.name" . }}
 {{- end }}
+
+{{- define "application.service-account-name" }}
```
d3adb5, line 72: This partial template (or whatever replaces it) needs to be used in `application/templates/rolebinding.yaml` as well and anywhere else where `template "application.name" $` (+ other branches) was used to come up with a service account name.

Would be good to add test cases for the templates that use it, though that might lead to duplicated tests.

d3adb5 (changes requested):
Left a few comments. I like the idea of adding a semantic value like this, but I feel the way it currently is introduces more confusion than clarity. I expanded upon this in one of the threads I opened below.

application/templates/_helpers.tpl
aslafy-z: I simplified this template a lot and used it anywhere it a service account was needed.

application/templates/_helpers.tpl
aslafy-z: the partial now only returns strings, do you think i should still quote the output?

application/templates/_helpers.tpl
```diff
@@ -68,3 +68,18 @@ reference:
...
+{{- if .Values.rbac.enabled }}
+  {{- if and .Values.rbac.serviceAccount.enabled .Values.rbac.existingServiceAccountName }}
+    {{- fail "Conflict: 'rbac.existingServiceAccountName' is set, but a new service account is being created. Please disable 'rbac.serviceAccount.enabled' or unset 'rbac.existingServiceAccountName'." }}
```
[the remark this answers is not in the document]
aslafy-z: I renamed `enabled` to `create` for it to be aligned with what is done in the community. I personally don't care about breaking changes if they are clearly indicated as such and major is bumped. I prefer to consistent and easily understandable APIs, over over-engineering for compatible solutions, even if it come with the cost of a breaking change. This is my personal take, I think it works for this chart. WDYT?

application/templates/cronjob.yaml
```diff
@@ -54,13 +54,7 @@ spec:
...
           {{- end }}
         spec:
-          {{- if $.Values.rbac.enabled }}
-          {{- if $.Values.rbac.serviceAccount.name }}
-          serviceAccountName: {{ $.Values.rbac.serviceAccount.name }}
-            {{- else }}
-          serviceAccountName: {{ template "application.name" $ }}
-          {{- end }}
-          {{- end }}
+          serviceAccountName: {{ include "application.serviceAccountName" $ }}
```
d3adb5, line 57: This may be another feature, but personally, I think we should add this:

```suggestion
          automountServiceAccountToken: {{ $.Values.rbac.enabled }}
          serviceAccountName: {{ include "application.serviceAccountName" $ }}
```

If RBAC is disabled, we shouldn't even allow mounting tokens for the default service account, I'd say. What do you think?

application/templates/deployment.yaml
```diff
@@ -74,6 +74,7 @@ spec:
           ]
 {{- end }}
     spec:
+      serviceAccountName: {{ include "application.serviceAccountName" $ }}
```
d3adb5, line 77: Same comment as in the other thread.

```suggestion
      automountServiceAccount: {{ $.Values.rbac.enabled }}
      serviceAccountName: {{ include "application.serviceAccountName" $ }}
```

application/templates/job.yaml
```diff
@@ -37,13 +37,7 @@ spec:
...
       {{- end }}
     spec:
-      {{- if $.Values.rbac.enabled }}
-      {{- if $.Values.rbac.serviceAccount.name }}
-      serviceAccountName: {{ $.Values.rbac.serviceAccount.name }}
-        {{- else }}
-      serviceAccountName: {{ template "application.name" $ }}
-      {{- end }}
-      {{- end }}
+      serviceAccountName: {{ include "application.serviceAccountName" $ }}
```
d3adb5, line 40: Same comment as in the other threads.

```suggestion
      automountServiceAccount: {{ $.Values.rbac.enabled }}
      serviceAccountName: {{ include "application.serviceAccountName" $ }}
```

d3adb5 (changes requested):
Final set of comments,  I promise. Reminder to add a unit test for it if you do take the suggestion. Otherwise, LGTM.

application/templates/_helpers.tpl
d3adb5: I think it's a good idea, yes, since the output is supposed to be a string.

rasheedamir:
@aslafy-z lets finalize and get this merged

application/templates/cronjob.yaml
aslafy-z: Sounds good to me,  that seems like a reasonable assumption.

karl-johan-grahn:
@aslafy-z Can you please resolve the conflicts?

d3adb5 (dismissed):
LGTM! Sorry for the delay in coming back to this.

aslafy-z:
@copilot resolve the merge conflicts in this pull request

aslafy-z:
@copilot resolve the merge conflicts in this pull request

PR #366, opened 2024-11-24
feat: add dnsPolicy support to Deployment

### LINKED ISSUE #365: Add dnsPolicy support
**Is your feature request related to a problem? Please describe.**
I'm using this chart for an application using hostNetworking (which is supported), so I would also like to configure dnsPolicy on the pod spec with ClusterFirstWithHostNet.

**Describe the solution you'd like**
Add support for dnsPolicy to Deployment (and perhaps job and cronjob, but those do not currently support dnsConfig either): a string property for the PodSpec.

**Describe alternatives you've considered**

**Additional context**

tomjo:
Fixes #365 

If wanted, I can also add it to job and cronjob but perhaps then support for dnsConfig should be added there as well?

### DIFF
--- a/application/values.yaml
+++ b/application/values.yaml
@@ -41,6 +41,11 @@ cronJob:
     #     requests:
     #         memory: 5Gi
     #         cpu: 1
+    #   dnsConfig:
+    #     options:
+    #     - name: ndots
+    #       value: '1'
+    #   dnsPolicy: ClusterFirst
 
 
 job:
@@ -71,6 +76,11 @@ job:
     #     requests:
     #         memory: 5Gi
     #         cpu: 1
+    #   dnsConfig:
+    #     options:
+    #     - name: ndots
+    #       value: '1'
+    #   dnsPolicy: ClusterFirst
 
 deployment:
   # -- (bool) Enable Deployment.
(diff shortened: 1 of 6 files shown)


aslafy-z:
I think adding `dnsConfig` too would be a nice addition. Also, I would prefer them to be added to deployments, jobs and cronjob at the same time. Can you do these changes?
Please use `make build-docs` to update the readme file.

application/values.yaml
```diff
@@ -251,6 +251,9 @@ deployment:
     # options:
     # - name: ndots
     #   value: '1'
+  # -- (string) DNS Policy.
+  # @section -- Deployment Parameters
+  dnsPolicy: ClusterFirst
```
aslafy-z, line 256: I would use the kubernetes default by default, by setting this value to an empty string (`""`).

application/values.yaml
```diff
@@ -251,6 +251,9 @@ deployment:
...
   # -- (object) Startup probe.
   # @default -- See below
```
aslafy-z, line 258: Please add dnsPolicy and dnsConfig to jobs and cronjobs sections.

aslafy-z (changes requested):

aslafy-z (approved):

rasheedamir (approved):

rasheedamir:
@aslafy-z you can merge it now

aslafy-z (approved):

PR #416, opened 2025-07-10
feat: add HTTPRoute resource for Gateway API

FacundoAcevedo:
Hi,
I'm trying to add support for the HTTPRoute resource to be able to use Gateway API

https://kubernetes.io/docs/concepts/services-networking/gateway/#api-kind-httproute

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,18 @@ helm delete --namespace test my-application
 | ingress.annotations | object | `nil` | Annotations for ingress. |
 | ingress.tls | list | `nil` | TLS configuration for ingress. Secrets must exist in the namespace. You may also configure Certificate resource to generate the secret. |
 
+### HTTPRoute Parameters
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| httpRoute.enabled | bool | `false` | Enable HTTPRoute (Gateway API). |
+| httpRoute.gatewayName | string | `"my-gateway"` | Name of the Gateway to attach this HTTPRoute to. |
+| httpRoute.gatewayNamespace | string | `""` | Namespace of the Gateway to attach this HTTPRoute to. If not set, the HTTPRoute will be attached to the Gateway in the same namespace as the HTTPRoute. |
+| httpRoute.hostnames | list | `["chart-example.local"]` | Hostnames for the HTTPRoute. |
+| httpRoute.additionalLabels | object | `{}` | Additional labels for HTTPRoute. |
+| httpRoute.annotations | object | `{}` | Annotations for HTTPRoute. |
+| httpRoute.rules | list | `[{"backendRefs":[{"name":"http","port":8080}],"matches":[{"path":{"type":"PathPrefix","value":"/"}}]}]` | Rules for HTTPRoute. |
+
 ### Route Parameters
 
 | Key | Type | Default | Description |
--- a/application/templates/httproute.yaml
+++ b/application/templates/httproute.yaml
@@ -1,0 +1,30 @@
+{{- if and (.Values.httpRoute).enabled -}}
+{{- if not (.Capabilities.APIVersions.Has "gateway.networking.k8s.io/v1") }}
+  {{- fail "There is no APIGateway resource definition in the target cluster!" }}
+{{- end }}
+---
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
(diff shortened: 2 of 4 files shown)


application/templates/httproute.yaml
```diff
@@ -0,0 +1,27 @@
...
+{{- if .Values.httpRoute.additionalLabels }}
+{{ toYaml .Values.httpRoute.additionalLabels | nindent 4 }}
+{{- end }}
+{{- if .Values.httpRoute.annotations }}
+  annotations:
+{{- toYaml .Values.httpRoute.annotations | nindent 4 }}
+{{- end }}
+spec:
+  parentRefs:
+    {{- toYaml .Values.httpRoute.parentRefs | nindent 4 }}
```
aslafy-z, line 22: I feel like fields should not be added to the template if they are not defined.

```suggestion
  {{- if .Values.httpRoute.parentRefs }}
  parentRefs:
    {{- toYaml .Values.httpRoute.parentRefs | nindent 4 }}
  {{- end }}
```

or

```suggestion
  {{- if .Values.httpRoute.parentRefs }}
  parentRefs: {{- toYaml .Values.httpRoute.parentRefs | nindent 4 }}
  {{- end }}
```

application/values.yaml
```diff
@@ -570,6 +570,40 @@ ingress:
...
+  # -- (list) Rules for HTTPRoute.
+  # @section -- HTTPRoute Parameters
+  rules:
+    - matches:
+        - path:
+            type: PathPrefix
+            value: /
+      backendRefs:
+        - name: http
+          port: 8080
```
aslafy-z, line 605: Can you please align this on the ingress implementation?

README.md
```diff
@@ -169,6 +169,18 @@ helm delete --namespace test my-application
...
 
+### HTTPRoute Parameters
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| httpRoute.enabled | bool | `false` | Enable HTTPRoute (Gateway API). |
+| httpRoute.gatewayName | string | `"my-gateway"` | Name of the Gateway to attach this HTTPRoute to. |
+| httpRoute.gatewayNamespace | string | `""` | Namespace of the Gateway to attach this HTTPRoute to. If not set, the HTTPRoute will be attached to the Gateway in the same namespace as the HTTPRoute. |
+| httpRoute.hostnames | list | `["chart-example.local"]` | Hostnames for the HTTPRoute. |
+| httpRoute.additionalLabels | object | `{}` | Additional labels for HTTPRoute. |
```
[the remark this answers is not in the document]
aslafy-z: In fact, #399 is not blocking for this PR to land. There's no change in the `application.labels` interface!

aslafy-z:
Please rebase onto the latest main! Thank you

aslafy-z (changes requested):
I left some comments. Please rebase onto the latest main. Thank you!

application/values.yaml
FacundoAcevedo: What do you mean by  align?

aslafy-z: I meant align so their values looks like the same. The goal is to auto-link this rule to the application service.

application/tests/httproute_test.yaml
```diff
@@ -0,0 +1,279 @@
...
+      - hasDocuments:
+          count: 0
+
+  - it: does not render when Gateway API is not available
+    set:
+      httpRoute:
+        enabled: true
+    asserts:
+      - failedTemplate:
+          errorMessage: "There is no APIGateway resource definition in the target cluster!"
```
d3adb5, line 21: Though [the documentation speaks of therre not being a native implementation,](https://kubernetes.io/docs/concepts/services-networking/gateway/#what-s-next) it's quite possible resources under the `gateway.networking.k8s.io` group might make their way to K8s core, just like Ingress resources. It might be good to add the current K8s version under capabilities here to ensure this test job doesn't break in the distant future, what do you think?


```suggestion
  - it: does not render when Gateway API is not available
    set:
      httpRoute:
        enabled: true
    capabilities:
      majorVersion: 1
      minorVersion: 34
    asserts:
      - failedTemplate:
          errorMessage: "There is no APIGateway resource definition in the target cluster!"
```

FacundoAcevedo: good catch

Issue #403, opened 2025-06-03
CronJob is missing startingDeadlineSeconds


kpxtaniey:
**Is your feature request related to a problem? Please describe.**
Current CronJob template is missing the option to add deadline for delayed Job start by setting `startingDeadlineSeconds`
**Describe the solution you'd like**
Extend CronJob template to allow optionally setting `startingDeadlineSeconds`.
**Describe alternatives you've considered**
None
**Additional context**
I've faced the issue that I am not able to configure it using the chart.

aslafy-z:
Closed by #420

PR #420, opened 2025-07-22
feat(cronjob): add startingDeadlineSeconds to job template

FacundoAcevedo:
Added startingDeadlineSeconds parameter to cronjobs as requested in issue https://github.com/stakater/application/issues/403

### DIFF
--- a/application/templates/cronjob.yaml
+++ b/application/templates/cronjob.yaml
@@ -39,6 +39,9 @@ spec:
 {{ end }}
   jobTemplate:
     spec:
+      {{- with $job.startingDeadlineSeconds }}
+      startingDeadlineSeconds: {{ . }}
+      {{- end }}
       {{- with $job.activeDeadlineSeconds }}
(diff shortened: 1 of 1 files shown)


FacundoAcevedo:
@aslafy-z is the file `values-test.yaml` ment to be used in manual testing? should I mention the key there too?

aslafy-z:
You can skip it it's ok.

application/templates/cronjob.yaml
```diff
@@ -39,6 +39,9 @@ spec:
 {{ end }}
   jobTemplate:
     spec:
+      {{- with $job.startingDeadlineSeconds }}
```
aslafy-z, line 42: I've just seen that a lot of integers are included with the with keyword. However, this would skip all 0 values.  Let's keep it like this in this PR, I will handle the change in a new one.

aslafy-z (approved):

PR #479, opened 2025-12-09
ci: set helm dry run to server side so it find vpa 

TehreemNisa:
helm lint is always client side , we can update the helm values to ignore ciMode in code

### DIFF
--- a/.github/workflows/pull_request.yaml
+++ b/.github/workflows/pull_request.yaml
@@ -43,7 +43,7 @@ jobs:
           var_file: ${{ env.CHART_NAME }}/values-test.yaml
       - name: Dry Run Chart
         run: |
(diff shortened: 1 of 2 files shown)


TehreemNisa:
> Hello, The error is still there, but shadowed:
> 
> ```
> level=INFO msg="funcMap fail" message="There is no VerticalPodAutoscaler resource definition in the target cluster!"
> ```

helm lint is always client side , but u can see dry-run succeeds now as i confirmed on the action-runners cluster VPA is installed and dry-run=server runs it on the parent cluster 

image

MuneebAijaz:
https://github.com/stakater/application/actions/runs/19540652867/job/55945764463 helm lint always had issues, because of helm lint not having the cluster context as @TehreemNisa mentioned. You can see in the older passing actions where VPA tests pass but helm lint logs the issue.

PR #576, opened 2026-07-09
feat: add Gateway API ListenerSet template

### LINKED ISSUE #547: Add ListenerSet support for cert‑manager Gateway API integration (alongside HTTPRoute)
**Description**  
The recent ingress‑nginx EOL announcement (Nov 26 2025) makes the Gateway API the main way to expose services. The chart already supports `HTTPRoute`, but not the newer `ListenerSet` resource that cert‑manager documents [here](https://cert-manager.io/docs/usage/gateway/#listenerset). Adding `ListenerSet` would let users define multiple listeners (TLS, HTTP, HTTPS) in a single object and simplify [migration to the Gateway API](https://cert-manager.io/announcements/2025/11/26/ingress-nginx-eol-and-gateway-api/).

**Proposed changes**

- **values.yaml:** add a `listenerSet` block (enabled `false` by default) with fields for name, port, protocol, and optional TLS secret.
- **templates:** render a `ListenerSet` CR when enabled and reference it from the `Gateway`.
- **RBAC:** conditionally grant permission to watch `listenersets`.
- **Docs & tests:** update README with usage examples and add Helm unit tests.

**Example values**

```yaml
listenerSet:
  enabled: true
  name: my-app-listeners
  listeners:
    - name: https
      port: 443
      protocol: HTTPS
      tls:
        secretName: {{ include "application.fullname" . }}-tls
    - name: http
      port: 80
      protocol: HTTP
```

**Benefits**

- Aligns the chart with the latest Gateway API best practices.  
- Reduces boilerplate for apps needing multiple listeners.  
- Provides a smooth migration path from ingress‑nginx.

**Request**  
Please add this optional `ListenerSet` support (e.g., in chart version 8.2.0) and update the documentation.

alex1989hu:
Adds a ListenerSet template so the chart can attach additional listeners to an existing Gateway, following the Gateway API ListenerSet resource.

Closes #547

### DIFF
--- a/application/templates/listenerset.yaml
+++ b/application/templates/listenerset.yaml
@@ -1,0 +1,22 @@
+{{- if and .Values.listenerSet .Values.listenerSet.enabled -}}
+{{- if not (.Capabilities.APIVersions.Has "gateway.networking.k8s.io/v1") }}
+{{- fail "There is no Gateway API resource definitions in the target cluster!" }}
+{{- end }}
+---
+apiVersion: gateway.networking.k8s.io/v1
+kind: ListenerSet
+metadata:
+  name: {{ template "application.name" . }}
+  namespace: {{ include "application.namespace" . }}
+  labels:
+    {{- include "application.labels" $ | nindent 4 }}
+    {{- if .Values.listenerSet.additionalLabels }}
+      {{ toYaml .Values.listenerSet.additionalLabels | nindent 4 }}
+    {{- end }}
+  {{- if .Values.listenerSet.annotations }}
+  annotations: {{- toYaml .Values.listenerSet.annotations | nindent 4 }}
+  {{- end }}
+spec:
+  listeners: {{- include "application.tplvalues.render" ( dict "value" .Values.listenerSet.listeners "context" $ ) | nindent 4 }}
+  parentRef: {{- include "application.tplvalues.render" ( dict "value" .Values.listenerSet.parentRef "context" $ ) | nindent 4 }}
+{{- end }}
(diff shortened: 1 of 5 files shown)


application/templates/listenerset.yaml
```diff
@@ -0,0 +1,22 @@
+{{- if and .Values.listenerSet .Values.listenerSet.enabled -}}
+{{- if not (.Capabilities.APIVersions.Has "gateway.networking.k8s.io/v1") }}
+  {{- fail "There is no APIGateway resource definition in the target cluster!" }}
```
aslafy-z:
```suggestion
  {{- fail "There is no Gateway API resource definitions in the target cluster!" }}
```

aslafy-z (changes requested):
Thank you for your contribution
Please address this small nit before we can proceed.

application/templates/listenerset.yaml
alex1989hu: I used what we have in `HTTPRoute` to be consistent: https://github.com/stakater/application/blob/ebcea70be307782d3eadc5081657be0d670ab415/application/templates/httproute.yaml#L3

Shall I modify both resources?

aslafy-z: I'm wondering if a good move wouldn't be to add the full group in the error message rather than a arbitrary name. What do you think ?
Just update it for your resource for now, and we will open a follow-up pr - if you have time, we welcome prs 🙃

alex1989hu: Understood. I've pushed the requested change (+ unit test update) https://github.com/stakater/application/commit/4563f80ce354866e67fe47295b1cc0c8f88d0a4e

aslafy-z (approved):

PR #603, opened 2026-08-05
fix(helpers): make backendRef port optional in HTTPRoute rules

### LINKED ISSUE #602: HTTPRoute template fails to render when backendRef port is omitted
**Describe the bug**
The `HTTPRoute` `backendRefs` template in `_helpers.tpl` requires `.port` to be set on every `backendRef` and fails the render if it's missing. Per the Gateway API spec, port is optional on a `BackendRef`, so this incorrectly forces users to always specify a port.

**To Reproduce**
1. Set `httpRoute.enabled: true` in values
2. Add a rule with a `backendRefs` entry that has no port field, e.g.:

```yaml
httpRoute:
  enabled: true
  parentRefs:
    - name: my-gateway
  hostnames:
    - example.com
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: /
      backendRefs:
        - name: example-service
```

3. Run helm template against the chart

**Expected behavior**
The chart renders successfully, omitting port from the manifest when it isn't provided.

**Actual behavior**
The template evaluates the missing `.port` as `0`, fails the `1–65535` range validation, and aborts the render with Invalid port value: 0. Port must be between `1` and `65535`.

alex1989hu:
The HTTPRoute backendRefs template required `.port` on every backendRef and failed the render if it was missing, but per the Gateway API spec port is optional on a `BackendRef` in general. It is only required when the referent is a core Kubernetes Service.

## What changed

- `port` is validated and rendered only when it is actually set, and left out of the manifest entirely when omitted (e.g. for `kind: ServiceImport` backends).
- Per the Gateway API spec, omitting `port` when the referent is a core Kubernetes Service (`group: ""` and `kind: Service`, the default) still fails the render, now with an explicit message: `backendRef "<name>": port is required when the referent is a Kubernetes Service`.
- A templated port that renders to an empty string (e.g. `port: '{{ .Values.optionalPort }}'` with the value unset) is treated as unset, so the optionality also works with template expressions.
- The `port` documentation in `values.yaml`, README, and `values.schema.json` now states when the field may be omitted.

## Note on the issue repro

The exact reproduction in #602 (no port, kind defaulting to Service) still fails by design, because the Gateway API spec requires port for Service referents and the HTTPRoute CRD would reject the manifest anyway. The difference is that the failure message now explains the actual constraint instead of `Invalid port value: 0`. Backends of any other kind render without a port as the issue expects.

## Tests

Unit tests cover: port omitted for a non-Service kind (rendered without port), port omitted for a Service (explicit failure), empty templated port for both cases, out-of-range and explicit-zero ports, mixed rules where only some backendRefs have a port, and a Service in a non-core group (port optional).

Closes #602

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ Please refer to the [Contributing Guide](CONTRIBUTING.md) for details on how to 
 | httpRoute.additionalLabels | object | `{}` | Additional labels for HTTPRoute. |
 | httpRoute.annotations | object | `{}` | Annotations for HTTPRoute. |
 | httpRoute.rules | list | `[{"backendRefs":[{"name":"{{ include \"application.name\" $ }}","port":"{{ (first $.Values.service.ports).port }}"}],"matches":[{"path":{"type":"PathPrefix","value":"/"}}]}]` | Rules for HTTPRoute. Keys and values are evaluated as templates. |
-| httpRoute.rules[0].backendRefs[0].port | int, tpl, null | `"{{ (first $.Values.service.ports).port }}"` | Port number or template expression for the backend service. |
+| httpRoute.rules[0].backendRefs[0].port | int, tpl, null | `"{{ (first $.Values.service.ports).port }}"` | Port number or template expression for the backend service. Required when the referent is a Kubernetes Service (the default kind); optional for other kinds such as ServiceImport, where it is omitted from the manifest when unset or when a template expression renders empty. |
 
 ### ListenerSet Parameters
 
--- a/application/templates/_helpers.tpl
+++ b/application/templates/_helpers.tpl
@@ -142,11 +142,11 @@ Usage:
   {{- range $rule.backendRefs }}
   {{- $group := .group | default "" }}
   {{- $kind := .kind | default "Service" }}
-  {{- $hasPort := not (kindIs "invalid" .port) }}
+  {{- /* A templated port may render to an empty string; treat it as unset. */}}
+  {{- $hasPort := not (or (kindIs "invalid" .port) (eq (toString .port) "")) }}
   {{- $isKubernetesService := and (eq $group "") (eq $kind "Service") }}
-  {{- $portVal := 0 }}
   {{- if $hasPort }}
-    {{- $portVal = .port | int }}
+    {{- $portVal := .port | int }}
     {{- if or (lt $portVal 1) (gt $portVal 65535) }}
       {{- fail (printf "Invalid port value: %v. Port must be between 1 and 65535" .port) }}
     {{- end }}
@@ -154,8 +154,8 @@ Usage:
     {{- fail (printf "backendRef %q: port is required when the referent is a Kubernetes Service" .name) }}
   {{- end }}
   - name: {{ .name }}
-    {{- if .port }}
-    port: {{ $portVal }}
+    {{- if $hasPort }}
+    port: {{ .port | int }}
     {{- end }}
     {{- if .weight }}
     weight: {{ .weight | int }}
--- a/application/templates/_helpers.tpl
+++ b/application/templates/_helpers.tpl
@@ -140,12 +140,15 @@ Usage:
   {{- if $rule.backendRefs }}
   backendRefs:
   {{- range $rule.backendRefs }}
+  {{- $kind := .kind | default "Service" }}
   {{- $portVal := 0 }}
   {{- if .port }}
     {{- $portVal = .port | int }}
     {{- if or (lt $portVal 1) (gt $portVal 65535) }}
       {{- fail (printf "Invalid port value: %v. Port must be between 1 and 65535" .port) }}
     {{- end }}
+  {{- else if eq $kind "Service" }}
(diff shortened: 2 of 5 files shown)


aslafy-z:
Thank you for your contribution @alex1989hu, please review my comment:

The new test case is not valid according to the Gateway API specification.

It defines:

```yaml
backendRefs:
  - name: example-service
```

Because neither `group` nor `kind` is specified, Gateway API infers the core API group and defaults `kind` to `Service`. This backendRef therefore refers to a Kubernetes Service, for which `port` is required:

https://gateway-api.sigs.k8s.io/reference/api-spec/main/spec/#:~:text=Port%20is%20required%20when%20the%20referent%20is%20a%20Kubernetes%20Service.

The chart does support explicit non-Service backends through the `group` and `kind` fields, but that is not what this test exercises.

As written, the PR allows an invalid Service backendRef to render successfully. I think the test should continue to require a port for the default Service case. Port omission should only be accepted when `group` or `kind` explicitly identifies a non-Service backend.

alex1989hu:
Do I understand correctly that you'd like to enforce the `port` field if the `kind` is omitted or set as `Service` ?

aslafy-z:
@alex1989hu I feel like it's the more correct thing to do, what do you think?

alex1989hu:
> @alex1989hu I feel like it's the more correct thing to do, what do you think?

I'm okay with that change. I've just pushed a commit [5f9dd47](https://github.com/stakater/application/pull/603/commits/5f9dd470c9e0603249c782eacadd7d5fdde07596) , could you please take a look?

aslafy-z:
Thanks, the Service/default-kind case is now handled correctly.

One remaining issue: `if .port` checks truthiness rather than whether the field was provided. For example, `port: 0` is falsey, so it bypasses the range validation. For a non-Service backend it would be silently omitted, while for a Service it would produce the less accurate "port is required" error.

Could we use `kindIs "invalid"` to distinguish an absent/null value from an explicitly provided falsey value, and reuse the same gate when rendering?

```gotemplate
{{- $group := .group | default "" }}
{{- $kind := .kind | default "Service" }}
{{- $hasPort := not (kindIs "invalid" .port) }}
{{- $isKubernetesService := and (eq $group "") (eq $kind "Service") }}
{{- $portVal := 0 }}

{{- if $hasPort }}
  {{- $portVal = .port | int }}
  {{- if or (lt $portVal 1) (gt $portVal 65535) }}
    {{- fail (printf "Invalid port value: %v. Port must be between 1 and 65535" .port) }}
  {{- end }}
{{- else if $isKubernetesService }}
  {{- fail (printf "backendRef %q: port is required when the referent is a Kubernetes Service" .name) }}
{{- end }}
```

Then use `$hasPort` instead of `if .port` when rendering the field.

The `group` check also avoids treating a custom resource such as `group: example.io, kind: Service` as a Kubernetes Service.

alex1989hu:
> Thanks, the Service/default-kind case is now handled correctly.
> 
> One remaining issue: `if .port` checks truthiness rather than whether the field was provided. For example, `port: 0` is falsey, so it bypasses the range validation. For a non-Service backend it would be silently omitted, while for a Service it would produce the less accurate "port is required" error.
> 
> Could we use `kindIs "invalid"` to distinguish an absent/null value from an explicitly provided falsey value, and reuse the same gate when rendering?
> 
> ```
> {{- $group := .group | default "" }}
> {{- $kind := .kind | default "Service" }}
> {{- $hasPort := not (kindIs "invalid" .port) }}
> {{- $isKubernetesService := and (eq $group "") (eq $kind "Service") }}
> {{- $portVal := 0 }}
> 
> {{- if $hasPort }}
>   {{- $portVal = .port | int }}
>   {{- if or (lt $portVal 1) (gt $portVal 65535) }}
>     {{- fail (printf "Invalid port value: %v. Port must be between 1 and 65535" .port) }}
>   {{- end }}
> {{- else if $isKubernetesService }}
>   {{- fail (printf "backendRef %q: port is required when the referent is a Kubernetes Service" .name) }}
> {{- end }}
> ```
> 
> Then use `$hasPort` instead of `if .port` when rendering the field.
> 
> The `group` check also avoids treating a custom resource such as `group: example.io, kind: Service` as a Kubernetes Service.

Good catch, thank you! I copied your proposal and added two tests in 8a06504a5b915bc322afd0927e62a8a08aa2881b

aslafy-z:
Thanks for the fix @alex1989hu!

I pushed a few additions on top of your commits before merging:

- A templated port that renders empty (e.g. port: '{{ .Values.optionalPort }}' with the value unset) is now treated as unset too, so the optionality composes with template expressions like the chart's own defaults use.
- The render guard now reuses the same presence check as the validation, instead of a separate truthiness test.
- Documented the "required for Service referents, optional otherwise" rule in values.yaml (README and schema regenerated).
- Tidied the new tests a bit (dropped a few duplicate cases) and added coverage for a Service in a non-core group.

Also rebased onto main. Merged, thanks again for the contribution!

PR #607, opened 2026-08-10
feat(servicemonitor): include additional service labels

### LINKED ISSUE #606: ServiceMonitor selector does not include service.additionalLabels
**Describe the bug**
When `service.additionalLabels` is set, those labels are applied to the `Service` object but are not included in the `ServiceMonitor`'s `spec.selector.matchLabels`. As a result, the generated `ServiceMonitor` can fail to select the `Service` it is meant to monitor whenever the additional labels are required for the selector to be unique or otherwise relied upon.

**To Reproduce**
1. Deploy the chart with the following values:
    ```yaml
    service:
      additionalLabels:
        custom: label
    serviceMonitor:
      enabled: true
      endpoints:
        - port: http
          interval: 30s
2. Run helm template on the chart.
3. Inspect the rendered `Service` manifest and note that `metadata.labels.custom: label` is present.
4. Inspect the rendered `ServiceMonitor` manifest and note that `spec.selector.matchLabels` does not contain custom: label.

**Expected behavior**
`spec.selector.matchLabels` on the `ServiceMonitor` should include the labels set under `service.additionalLabels`, so the `ServiceMonitor` selector stays consistent with the labels actually applied to the `Service`.

**Actual behavior**
`spec.selector.matchLabels` only contains the labels from `application.labels`; `service.additionalLabels` are silently ignored, so the `ServiceMonitor` may not match the `Service` as intended or match another services (in my case, an additional `headless` `Service`).

alex1989hu:
Closes #606

### DIFF
--- a/application/templates/servicemonitor.yaml
+++ b/application/templates/servicemonitor.yaml
@@ -20,6 +20,9 @@ spec:
   selector:
     matchLabels:
 {{ include "application.labels" $ | indent 6 }}
+{{- if .Values.service.additionalLabels }}
+{{- toYaml .Values.service.additionalLabels | nindent 6 }}
+{{- end }}
   namespaceSelector:
     matchNames:
     - {{ include "application.namespace" . }}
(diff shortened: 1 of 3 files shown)


aslafy-z (approved):
LGTM! Thank you @alex1989hu!

PR #620, opened 2026-08-21
feat: Support for StatefulSets, DaemonSets and Headless Services

### LINKED ISSUE #319: Deployment Kind override
**Is your feature request related to a problem? Please describe.**
Some time there is a need of overriding `deployment.kind` to something like `StatefulSet`

**Describe the solution you'd like**
Something like `deployment.kind` to be set using chart values

**Describe alternatives you've considered**
nothing

**Additional context**

davidkim83:
### Changes
- enhanced the deployment template to accommodate variations between `Deployment`/`StatefulSet`/`DaemonSet`
- created a new template for headless `Service` to accompany `StatefulSet`
- HPA and VPA now reflects the `deployment.kind`
- added/edited test cases to cover the above changes
- updated docs to increase scope of chart (no longer 'stateless-only')

### Closes
#319 

### Remarks
- The idea of changing the top attribute from `deployment` to `workload` was discarded as it is deemed too intrusive.

### DIFF
--- a/application/templates/headlessservice.yaml
+++ b/application/templates/headlessservice.yaml
@@ -17,7 +17,9 @@ metadata:
 spec:
   clusterIP: None
   type: ClusterIP
+  {{- if .Values.headlessService.publishNotReadyAddresses }}
   publishNotReadyAddresses: {{ .Values.headlessService.publishNotReadyAddresses | default false }}
+  {{- end }}
   selector:
 {{ include "application.selectorLabels" . | indent 4 }}
   ports:
--- a/application/templates/headlessservice.yaml
+++ b/application/templates/headlessservice.yaml
@@ -1,0 +1,24 @@
+{{- if (.Values.headlessService).enabled }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ .Values.headlessService.name | default (printf "%s-headless" (include "application.name" .)) }}
+  namespace: {{ template "application.namespace" $ }}
+  labels:
+  {{- include "application.labels" $ | nindent 4 }}
+  {{- if .Values.headlessService.additionalLabels }}
+    {{- toYaml .Values.headlessService.additionalLabels | nindent 4 }}
+  {{- end }}
+  {{- if .Values.headlessService.annotations }}
+  annotations:
+    {{- include "application.tplvalues.render" ( dict "value" .Values.headlessService.annotations  "context" $ ) | nindent 4 }}
+{{- end }}
+spec:
+  clusterIP: None
+  type: ClusterIP
+  selector:
+{{ include "application.selectorLabels" . | indent 4 }}
+  ports:
+{{ toYaml .Values.headlessService.ports | indent 4 }}
+{{- end }}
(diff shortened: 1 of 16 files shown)


rasheedamir:
Thanks for the PR @davidkim83 

@aslafy-z what is your opinion on supporting statefulsets, daemonsets, etc. in this chart? the original idea with this chart has been for stateless applications only

davidkim83:
@rasheedamir my pleasure!

Please check [this comment](https://github.com/stakater/application/issues/319#issuecomment-3405128431) where @aslafy-z already gave his opinion, but I won't stop you from reconfirming :)

FWIW, this chart already supports Jobs & CronJobs. It would gain so much potential from supporting other types of long-running workloads.

I had also thought of separate top-level attributes (such as `statefulset:` and `daemonset:`, but then the chart would no longer be application, but application**s**. I think it's good that the chart remains limited to deploying a _singular_ workload, and not many.

application/templates/headlessservice.yaml
```diff
@@ -0,0 +1,24 @@
...
+  labels:
+  {{- include "application.labels" $ | nindent 4 }}
+  {{- if .Values.headlessService.additionalLabels }}
+    {{- toYaml .Values.headlessService.additionalLabels | nindent 4 }}
+  {{- end }}
+  {{- if .Values.headlessService.annotations }}
+  annotations:
+    {{- include "application.tplvalues.render" ( dict "value" .Values.headlessService.annotations  "context" $ ) | nindent 4 }}
+{{- end }}
+spec:
```
alex1989hu, line 17: I would add `spec.publishNotReadyAddresses` field which I typically need when I use `headless` service. Thouhgts?

davidkim83: No objections, would this work?
- Add `publishNotReadyAddresses` in the headless service template. Set to `false` as default in the template since that is Kubernetes Service spec's default
- Add an empty attribute with the same name in the `headlessService` section in the `values.yaml` file

davidkim83: Done

application/values.yaml
```diff
@@ -611,6 +644,33 @@ service:
...
+  # -- (list) Ports for applications service.
+  # @section -- Headless Service Parameters
+  ports:
+    - port: 8080
+      name: http
+      protocol: TCP
+      # -- (int, string, null) Target port on pods. Accepts port number or port name (IANA_SVC_NAME).
+      # @section -- Headless Service Parameters
+      targetPort: 8080
+  # -- (bool) Propagate SRV DNS records for Pods to enable peer discovery, by disregarding any indications of ready/not-ready for Pods.
```
alex1989hu: Based on the current values, I assume the correct type is `(bool, null)`

application/templates/headlessservice.yaml
alex1989hu: I would set the default value to false in the `values.yaml`, but I might be wrong. By the way, I can't resolve this thread.

davidkim83: Resolved it by checking once more if the value exists

application/values.yaml
davidkim83: Fixed!

PR #626, opened 2026-08-28
feat: Support Istio Virtual Service

LeoDOD:
## Changes
- added `virtualService` section to `values.yaml` with full documentation (`hosts`, `gateways`, `exportTo`, `http`, `tls`, `tcp`, `routes`)
- added `virtualService` schema block to `values.schema.json`
- created `templates/virtualservice.yaml` : supports single-VS and multi-VS (`routes` map) modes; `http`/`tls`/`tcp` are `toYaml` pass-through.
- API version auto-detects `networking.istio.io/v1` vs `v1beta1`
- Added test to validate the new template.

## Remarks
- `virtualService.enabled` defaults to `false` to be fully backwards compatible
- when `routes` is omitted, a single VirtualService is rendered using the top-level `virtualService` fields; when `routes` is a map, one VirtualService is rendered per entry (name suffix skipped when key is `"default"`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,28 +2,22 @@
 
 # Application
 
-Generic Helm chart for deploying stateless applications on Kubernetes. Supports Deployments, Jobs, and CronJobs along with common companion resources (Services, Ingress, RBAC, autoscaling, monitoring, certificates, and more).
+Generic helm chart for applications which:
+
+- are stateless
+- creates only namespace scoped resources (e.g. it doesn't need CRB - Cluster Role Bindings)
+- don't need privileged containers
+- don't call the underlying Kubernetes API or use the underlying etcd as a database by defining custom resources
+- run either as deployment, job or cronjob
 
 ## Installing the Chart
 
-### From the OCI registry
-
-```shell
-helm install my-application oci://ghcr.io/stakater/charts/application --namespace test
-# or, to install a specific version
-helm install my-application oci://ghcr.io/stakater/charts/application --version <version> --namespace test
-```
-
-### From the Helm repository (deprecated)
-
-> **Note:** The Helm repository is deprecated in favor of the OCI registry above.
+To install the chart with the release name `my-application` in namespace `test`:
 
 ```shell
 helm repo add stakater https://stakater.github.io/stakater-charts
 helm repo update
 helm install my-application stakater/application --namespace test
-# or, to install a specific version
-helm install my-application stakater/application --version <version> --namespace test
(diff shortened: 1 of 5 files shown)


aslafy-z:
Thanks for the PR, the template and tests are in good shape. Three things block merge, the rest are suggestions.

## Blocking

**1. Wildcard hosts render invalid YAML.** `templates/virtualservice.yaml` emits `hosts`, `gateways` and `exportTo` entries as `- {{ . }}` unquoted, so `*.example.com` (the exact usage documented in `values.yaml`) and `exportTo: ["*"]` will fail. Please switch to `{{ . | quote }}` or render the lists with `toYaml`, and add a test case with a wildcard host and `exportTo: ["*"]`.

**2. Feature is not enabled in CI values.** Per CONTRIBUTING, new features must be enabled in `ci/values.yaml` or `ci/values-crds.yaml` so the snapshot and render jobs cover them. Please add an enabled `virtualService` example (including the `routes` map mode) to `ci/values-crds.yaml`, add `networking.istio.io/v1` and `networking.istio.io/v1/VirtualService` to the `crds` profile `api-versions` list in `.github/workflows/ci.yaml` and to the capabilities list in `tests/common_test.yaml`, then run `mise run test -u` and commit the updated snapshot.

**3. Labels and annotations.** Every other networking template (`ingress.yaml`, `httproute.yaml`) supports `additionalLabels` and `annotations`. Please add `virtualService.additionalLabels` and `virtualService.annotations` (and the per-route equivalents) using the same pattern, then regenerate README and schema.

## Suggestions

- **apiVersion fallback.** Other CRD templates either gate on `Capabilities.APIVersions.Has` or `fail` when the CRD is absent (see `httproute.yaml`). This template silently emits `v1beta1` whenever the CRD is not present, which is what `helm template` and GitOps renders always get. Either adopt the `fail` pattern or document in the values comment that offline rendering always yields `v1beta1`.
- **Require `hosts`.** `enabled: true` with defaults renders an empty `spec:`. A `required "virtualService.hosts is required"` per route would give a clearer error.
- **Name length in `routes` mode.** The `-<key>` suffix is appended after `application.name` is already truncated to 63 characters, so long names overflow. Wrap the result in `trunc 63 | trimSuffix "-"`.
- **Schema typing.** `exportTo` is `(list, null)` while `hosts` and `gateways` are `(list)` with the same `[]` default. Align them.

Happy to re-review once the two blocking items are in.

